PR #20, opened 2018-01-12
[FOGL-917] Added the ability to enable/disable schedule

monika-Dianomic:


src/app/scheduler/scheduled-process/scheduled-process.component.html
```diff
@@ -59,6 +59,10 @@
                 <i class="fa fa-minus-circle" aria-hidden="true"></i>
               </div>
             </td>
+            <td>
+              <button *ngIf='schedule.enabled' class="button is-text" (click)="disableSchedule(schedule.id)">disable</button>
```
mshariq-nerd: can't we use `toggle button` instead of these two separate buttons for enable and disable schedule? IMHO that will look good on GUI. cc/ @praveen-garg

src/app/scheduler/scheduled-process/scheduled-process.component.ts
```diff
@@ -156,4 +156,66 @@ export class ScheduledProcessComponent implements OnInit {
   getISODay(index: number) {
     return weekDays[index];
   }
+
+  /**
+   * Disable schedule
+   * @param schedule_id id of the schedule to disable
+   */
+  public disableSchedule(schedule_id) {
+    console.log('Schedule ID:', schedule_id);
```
mshariq-nerd: remove this log

src/app/scheduler/scheduled-process/scheduled-process.component.html
mshariq-nerd: or if still we want to use this approach we can handle it with one button on  ternary operator condition check.

praveen-garg: I think simple is better!  better readability!

re: switch on/ off  let's not add 1 more dependency for looks!

src/app/scheduler/scheduled-process/scheduled-process.component.ts
praveen-garg: no, but say console.log('Disabling Schedule :', schedule_id);

do the same for enable!

PR #21, opened 2018-01-15
Added alert for enable/disable schedule and changed colour for disabled schedule

monika-Dianomic:
Added alert for enable/disable schedule and changed colour for disabled schedule

### DIFF
--- a/src/app/modal/modal.component.html
+++ b/src/app/modal/modal.component.html
@@ -2,10 +2,10 @@
   <div class="modal-background"></div>
   <div class="modal-card">
     <section class="modal-card-body">
-      Are you sure, You want to delete the schedule <span class="has-text-black has-text-weight-semibold">{{childData.name}}?</span>
+      {{childData.message}}<span class="has-text-black has-text-weight-semibold">&nbsp;{{childData.name}}?</span>
     </section>
     <footer class="modal-card-foot">
-      <button class="button is-success" (click)="delete()">Yes</button>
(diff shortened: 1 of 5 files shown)


src/app/scheduler/scheduled-process/scheduled-process.component.html
```diff
@@ -30,7 +30,7 @@
               <th></th>
             </tr>
           </thead>
-          <tr *ngFor="let schedule of scheduleData">
+          <tr *ngFor="let schedule of scheduleData" class="{{(!schedule.enabled) ? 'fade' : ''}}">
```
mshariq-nerd: We must use angular way to apply css class dynamically. use `NgClass` instead of simple `class` condition

src/app/modal/modal.component.html
```diff
@@ -2,10 +2,10 @@
   <div class="modal-background"></div>
   <div class="modal-card">
     <section class="modal-card-body">
-      Are you sure, You want to delete the schedule <span class="has-text-black has-text-weight-semibold">{{childData.name}}?</span>
+      {{childData.message}}<span class="has-text-black has-text-weight-semibold">&nbsp;{{childData.name}}?</span>
     </section>
     <footer class="modal-card-foot">
-      <button class="button is-success" (click)="delete()">Yes</button>
+      <button class="button is-success" (click)="callMethod()">Yes</button>
```
praveen-garg: name it better! may be `triggerAction()`

PR #22, opened 2018-01-15
Fix get of all categories to match spec and latest code

MarkRiddoch:


praveen-garg:
@MarkRiddoch this fix is already in develop. FogLAMP master and 1.1 branch has endpoint /categories itself hence the same in corresponding gui branch.

https://github.com/foglamp/FogLAMP/blob/1.1/python/foglamp/services/core/api/configuration.py#L18
https://github.com/foglamp/FogLAMP/blob/master/python/foglamp/services/core/api/configuration.py#L18

PR #27, opened 2018-01-19
Fixed configuration wrong value save

mshariq-nerd:


src/app/configuration-manager/configuration-manager.component.ts
```diff
@@ -20,24 +20,22 @@ export class ConfigurationManagerComponent implements OnInit {
     this.configService.getCategories().
       subscribe(
       data => {
-         /** request completed */
-         this.ngProgress.done();
-         
-         if (data.error) {
```
praveen-garg: NO! master rest api return error payload with 200 status code itself

PR #33, opened 2018-01-23
(FOGL-948) Allowed to start schedule of manual type

monika-Dianomic:
Allowed to start schedule of manual type.
NOTE: Need a method on server side to return the status (Schedule started or not) of Schedule.


praveen-garg:
blocked by FOGL-987: Missing key (whether schedule is running or not) in json response for Manual schedule

PR #47, opened 2018-02-15
Added checkmark to enable/disable schedule

monika-Dianomic:
Added checkmark to enable/disable schedule while creating the schedule.


praveen-garg:
I think we should keep this operation explicit and let `create schedule` handle  default state.

PR #66, opened 2018-03-23
FOGL-528: User interface for backup/restore

monika-Dianomic:
- [x] tested with https + auth mandatory
- [x] tested with no backup exists and other failures simulation

**NOTE:** Restore backup has an issue on API side **(FOGL-834)**

### DIFF
--- a/src/app/backup-restore/backup-restore.component.ts
+++ b/src/app/backup-restore/backup-restore.component.ts
@@ -11,7 +11,7 @@ import { NgProgress } from 'ngx-progressbar';
 export class BackupRestoreComponent implements OnInit {
   public backupData = [];
 
-  // Object to hold data of certificate to delete
+  // Object to hold child data
   public childData = {
     id: '',
     name: '',
--- a/src/app/backup-restore/backup-restore.component.ts
+++ b/src/app/backup-restore/backup-restore.component.ts
@@ -67,7 +67,7 @@ export class BackupRestoreComponent implements OnInit {
     subscribe(
       data => {
         this.ngProgress.done();
-        this.alertService.success("Backup created successfully");
+        this.alertService.success(data.message);
         this.getBackup();
       },
       error => {
@@ -87,7 +87,7 @@ export class BackupRestoreComponent implements OnInit {
     subscribe(
       data => {
         this.ngProgress.done();
-        this.alertService.success("Restored successfully");
+        this.alertService.success(data.message);
         this.getBackup();
       },
       error => {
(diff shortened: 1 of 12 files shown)


src/app/backup-restore/backup-restore.component.ts
```diff
@@ -0,0 +1,123 @@
...
+    );
+  }
+
+  public requestBackup() {
+    this.ngProgress.start();
+    this.backupRestoreService.requestBackup().
+    subscribe(
+      data => {
+        this.ngProgress.done();
+        this.alertService.success("Backup created successfully");
```
praveen-garg: I think  we should display whatever message is returned from api

src/app/backup-restore/backup-restore.component.ts
```diff
@@ -0,0 +1,123 @@
...
+    );
+  }
+
+  public restoreBackup(id) {
+    this.ngProgress.start();
+    this.backupRestoreService.restoreBackup(id).
+    subscribe(
+      data => {
+        this.ngProgress.done();
+        this.alertService.success("Restored successfully");
```
praveen-garg: I think  we should display whatever message is returned from api

src/app/backup-restore/backup-restore.component.ts
```diff
@@ -0,0 +1,123 @@
...
+    );
+  }
+
+  public deleteBackup(id) {
+    this.ngProgress.start();
+    this.backupRestoreService.deleteBackup(id).
+    subscribe(
+      data => {
+        this.ngProgress.done();
+        this.alertService.success("Backup deleted successfully");
```
praveen-garg: I think  we should display whatever message is returned from api

src/app/backup-restore/backup-restore.component.ts
monika-Dianomic: It is returning `{"status": "running"}`

src/app/backup-restore/backup-restore.component.ts
monika-Dianomic: It is returning `{"status": "running"}`

src/app/backup-restore/backup-restore.component.ts
praveen-garg: Yes please, if needs to be fixed, it should be on API side

src/app/backup-restore/backup-restore.component.ts
praveen-garg: Yes please, if needs to be fixed, it should be on API side

src/app/services/backup-restore.service.ts
```diff
@@ -0,0 +1,48 @@
+import { Injectable } from '@angular/core';
+import { HttpClient } from '@angular/common/http';
+import { Observable } from 'rxjs/Rx';
```
mshariq-nerd: `import { Observable } from 'rxjs/Observable' instead of importing full library.

PR #74, opened 2018-03-27
change in build.sh

mshariq-nerd:
build.sh with help, version usage info; and license etc

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -e
 
 __author__="Mohd. Shariq"
 __copyright__="Copyright (c) 2017 OSIsoft, LLC"
@@ -48,8 +49,12 @@ remove_unwanted_files () {
     rm -rf .DS_Store
 }
 
+compress_bundle_js_files () {
+    gzip --keep dist/*.bundle.js
+}
+
 show_files_with_size () {
-    for f in dist/*.bundle*; do
+    for f in dist/*.*; do
         FILE_SIZE=$(du -h "$f")
         echo "${FILE_SIZE}"
     done
@@ -57,10 +62,10 @@ show_files_with_size () {
 
 build () {
     
-    echo -e  INFO: "${CINFO} Cleaning the dependencies ... ${CRESET}"
+    echo -e  INFO: "${CINFO} Cleaning the build and dependencies ... ${CRESET}"
     yarn clean
 
-    echo -e  INFO: "${CINFO} Adding dependencies ... ${CRESET}"
+    echo -e  INFO: "${CINFO} Installing dependencies ... ${CRESET}"
     yarn install
 
     echo -e  INFO: "${CINFO} Creating production build ... ${CRESET}"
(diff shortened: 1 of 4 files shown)


build.sh
```diff
@@ -20,35 +54,82 @@ show_files_with_size () {
...
+    echo "${USAGE}"
+
+  elif [ "$OPTION" == "VERSION" ]
+  then
+    echo $__version__
+
+  elif [ "$OPTION" == "START" ]
+  then
+    build
+
```
ashish-jabble: Remove `START`?

build.sh
```diff
@@ -20,35 +54,82 @@ show_files_with_size () {
...
-RELEASBLE_BUILD="foglamp-gui-${VERSION}.tar.gz"
-echo "Creating compressed build artifacts for release ..." 
-tar -zcvf ${RELEASBLE_BUILD} dist &>/dev/null
-echo "Created ${RELEASBLE_BUILD}"
-echo "Done."
+    execute_command
+  done
+else
+  OPTION="START"
+  execute_command
```
ashish-jabble: else ?

build.sh
```diff
@@ -20,35 +54,82 @@ show_files_with_size () {
...
+    T_SIZE=$(du -hs dist)
+    echo -e INFO " ${CWARN} Size: ${T_SIZE} ${CRESET}" 
+
+    echo -e INFO: "${CINFO} Removing unwanted contents ... ${CRESET}"
+    remove_unwanted_files
+
+    T_SIZE=$(du -hs dist)
+    echo -e INFO: "${CINFO} Deployable dist size ${CWARN}  ${T_SIZE} ${CRESET}" 
+
+    RELEASBLE_BUILD="foglamp-gui-${__version__}.tar.gz"
```
ashish-jabble: `RELEASABLE_BUILD`

build.sh
praveen-garg: when run without args, start building

build.sh
praveen-garg: no, we are setting this as default option

build.sh
praveen-garg: but yes, we can simply call `build` here and remove `elif [ "$OPTION" == "START" ]` block /cc @mshariq-nerd

build.sh
```diff
@@ -99,8 +107,7 @@ execute_command() {
   then
     echo $__version__
 
-  elif [ "$OPTION" == "START" ]
-  then
+  else
```
praveen-garg: keep this only with if / elif checks

build.sh
```diff
@@ -130,6 +137,5 @@ then
     execute_command
   done
 else
-  OPTION="START"
   execute_command
```
praveen-garg: call build here not as another else  in execute_command

nginx.conf
```diff
@@ -13,25 +13,29 @@ http {
...
         location / {
             try_files $uri$args $uri$args/ /index.html;
         }
 
-        # serve static content
-        location /static/ {
-            # path for static files
-            root .;
-        }
+        # remove this?!
```
praveen-garg: I think we should remove this or find a way for dynamic proxy?
As we can have call to 0.0.0.0:1995 via https or as per configured port

mshariq-nerd:
fixed per feedback.

PR #77, opened 2018-03-28
FOGL-1202: Day dropdown field set to blank for timed schedule

monika-Dianomic:
Day dropdown field set to blank for timed schedule

### DIFF
--- a/src/app/scheduler/scheduled-process/scheduled-process.component.ts
+++ b/src/app/scheduler/scheduled-process/scheduled-process.component.ts
@@ -7,7 +7,6 @@ import { CreateScheduleComponent } from '../create-schedule/create-schedule.comp
 import { NgProgress } from 'ngx-progressbar';
 
 enum weekDays {
-  Select = 0,
   Mon = 1,
   Tue = 2,
   Wed = 3,
--- a/src/app/scheduler/scheduled-process/scheduled-process.component.ts
+++ b/src/app/scheduler/scheduled-process/scheduled-process.component.ts
@@ -7,6 +7,7 @@ import { CreateScheduleComponent } from '../create-schedule/create-schedule.comp
 import { NgProgress } from 'ngx-progressbar';
 
 enum weekDays {
+  Select = 0,
   Mon = 1,
   Tue = 2,
   Wed = 3,
--- a/src/app/services-health/services-health.component.html
+++ b/src/app/services-health/services-health.component.html
@@ -13,7 +13,7 @@
             </button>
           </h5>
           <h6 class='subtitle is-6'>
-            <small>updated: {{time|date:'MMM d, y HH:mm:ss'}}</small>
+            <small>Last updated: {{time|date:'MMM d, y HH:mm:ss'}}</small>
           </h6>
         </div>
       </div>
--- a/src/app/update-modal/update-modal.component.ts
+++ b/src/app/update-modal/update-modal.component.ts
@@ -64,13 +64,13 @@ export class UpdateModalComponent implements OnInit, OnChanges {
    * getSelectedDay
    */
   public getSelectedDay(index) {
-    let selected_day;
+    let selectedDay;
     if (index == null) {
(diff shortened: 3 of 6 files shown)


src/app/update-modal/update-modal.component.ts
```diff
@@ -64,7 +64,12 @@ export class UpdateModalComponent implements OnInit, OnChanges {
    * getSelectedDay
    */
   public getSelectedDay(index) {
-    let selected_day = this.days[index - 1];
+    let selected_day;
```
praveen-garg: selectedDay ?

src/app/update-modal/update-modal.component.ts
```diff
@@ -140,7 +145,11 @@ export class UpdateModalComponent implements OnInit, OnChanges {
     if (this.form.get('type').value == '2') {   // If Type is TIMED == 2
       time = Utils.convertTimeToSec(this.form.get('time').value);
       let index = this.form.get('day').value != undefined ? this.days.indexOf(this.form.get('day').value) : 0;
-      this.form.controls['day'].setValue(index + 1);
+      if (index == 0) {
```
praveen-garg: === ?

src/app/scheduler/scheduled-process/scheduled-process.component.ts
```diff
@@ -7,6 +7,7 @@ import { CreateScheduleComponent } from '../create-schedule/create-schedule.comp
 import { NgProgress } from 'ngx-progressbar';
 
 enum weekDays {
+  Select = 0,
```
praveen-garg: can not we avoid this or we used this for direct binding?

src/app/update-modal/update-modal.component.ts
```diff
@@ -64,7 +64,12 @@ export class UpdateModalComponent implements OnInit, OnChanges {
    * getSelectedDay
    */
   public getSelectedDay(index) {
-    let selected_day = this.days[index - 1];
+    let selected_day;
+    if (index == null) {
+      selected_day = 'Select';
```
praveen-garg: if we have weekDays enum with Select should not we use this.days[0] ?

src/app/scheduler/scheduled-process/scheduled-process.component.ts
praveen-garg: this is affecting create schedule too, so let's remove it and it does not make sense in `weekDays`

Also in Dropdown default should be `None` instead of `Select`

src/app/services-health/services-health.component.html
```diff
@@ -13,7 +13,7 @@ <h5 class="title is-5">
             </button>
           </h5>
           <h6 class='subtitle is-6'>
-            <small>updated: {{time|date:'MMM d, y HH:mm:ss'}}</small>
+            <small>Last updated: {{time|date:'MMM d, y HH:mm:ss'}}</small>
```
praveen-garg: thanks for this! 👍

src/app/update-modal/update-modal.component.ts
```diff
@@ -139,8 +144,12 @@ export class UpdateModalComponent implements OnInit, OnChanges {
     let time; 
     if (this.form.get('type').value == '2') {   // If Type is TIMED == 2
       time = Utils.convertTimeToSec(this.form.get('time').value);
-      let index = this.form.get('day').value != undefined ? this.days.indexOf(this.form.get('day').value) : 0;
-      this.form.controls['day'].setValue(index + 1);
+      let index = this.form.get('day').value != undefined ? (this.days.indexOf(this.form.get('day').value)+1) : 0;
```
praveen-garg: why +1 now?

monika-Dianomic: `indexOf()` returns index, which is starting from 0 but in our `weekDays` enum, index starts from 1. To match the index values, it required to use `+1` here.

Issue #73, opened 2018-03-27
create a deploy.sh with --build-url option


praveen-garg:

praveen-garg:
`./deploy`

cat: package.json: No such file or directory

PR #79, opened 2018-03-28
Build deployment script 

mshariq-nerd:
#73

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -18,66 +18,66 @@ CWARN="${CPFX}0;33m"
 FOGLAMP_GUI_VER=1.2.0
 NGINX_STABLE_VER=1.12.2
 
-# Define Variable tecreset
+
 tecreset=$(tput sgr0)
 
-conection_details() {
-    # Check OS Type
-    os=$(uname -o)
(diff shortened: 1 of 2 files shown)


deploy.sh
```diff
@@ -0,0 +1,107 @@
...
+OPTIONS
+  Multiple commands can be specified but they all must be
+  specified separately (-hv is not supported).
+
+  -h, --help     Display this help text
+  -v, --version  Display this script's version information
+  -s, --start    Install and start
+
+EXIT STATUS
+  This script exits with status code 1 when errors occur.
```
praveen-garg: no; not sure where we used exit 1;;

praveen-garg: also for build.sh

deploy.sh
```diff
@@ -0,0 +1,107 @@
...
+DESCRIPTION
+  This script installs foglamp-gui with nginx-light
+
+OPTIONS
+  Multiple commands can be specified but they all must be
+  specified separately (-hv is not supported).
+
+  -h, --help     Display this help text
+  -v, --version  Display this script's version information
+  -s, --start    Install and start
```
praveen-garg: default should be start; if we want to install and start supply `--install --start` `-i -s` ?

PR #82, opened 2018-04-03
fix 404 when manual url or refresh is tried with default nginx conf

praveen-garg:
see https://github.com/angular/angular-cli/issues/5113


praveen-garg:
Another option is to use provided nginx conf. with

```
location / {
            try_files $uri$args $uri$args/ /index.html;
        }
```

Or edit default conf with this

PR #88, opened 2018-04-09
Fix assorted issues 

mshariq-nerd:
## Issue fixed
*Reset password issues* : 
1. After expiration of Password, “Your password has been expired. Please set your password again” displayed on UI also along with change password screen. 
2. Confirm password alert is not showing
3. Current password grey text should be displayed instead of old password for consistency
4. For Super Admin, All controls i.e. change role, reset password, delete and logout should be hide to another Admin user.

*Ping issues*:
1. Click on login button, Page is redirecting on Dashboard instead of login page
2. If default ping setting is 5 second, Default “Every 5 Seconds” value should be selected and displayed

Yes

## Checklist


src/app/services/ping.service.ts
```diff
@@ -1,8 +1,16 @@
 import { Injectable } from '@angular/core';
 import { BehaviorSubject } from 'rxjs/BehaviorSubject';
+import { POLLING_INTERVAL } from '../utils';
 
 @Injectable()
 export class PingService {
-  isPingIntervalChanged: BehaviorSubject<boolean> = new BehaviorSubject(false);
+  isPingIntervalChanged: BehaviorSubject<number> = new BehaviorSubject(0);
```
praveen-garg: isPingIntervalChanged > PingIntervalChanged as it not boolean now

praveen-garg:
@mshariq-nerd  we need to update changelog for this settings option

PR #90, opened 2018-04-11
Fixed protocol issue with service shutdown

monika-Dianomic:
## What do these changes do?

Fixed protocol issue with service shutdown.
Now, on service Health page service 'shutdown' works on http protocol of the given service.

No

## Related issue number

None

## Checklist

- [x] Code is well written w.r.t. best practices and proper comments

### DIFF
--- a/src/app/modal/modal.component.ts
+++ b/src/app/modal/modal.component.ts
@@ -7,7 +7,8 @@ import { SchedulesService, UserService, AlertService } from '../services/index';
   templateUrl: './modal.component.html'
 })
 export class ModalComponent implements OnInit {
-  @Input() childData: { id: Number, name: any, key: any, message: any, protocol: any };
+  @Input() childData: { id: Number, name: any, key: any, message: any };
+  @Input() shutDownServiceData: { port: Number, key: any, message: any, protocol: string };
   @Output() enable = new EventEmitter<Number>();
   @Output() disable = new EventEmitter<Number>();
   @Output() delete = new EventEmitter<Number>();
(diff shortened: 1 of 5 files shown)


src/app/modal/modal.component.ts
```diff
@@ -7,11 +7,11 @@ import { SchedulesService, UserService, AlertService } from '../services/index';
   templateUrl: './modal.component.html'
 })
 export class ModalComponent implements OnInit {
-  @Input() childData: { id: Number, name: any, key: any, message: any };
+  @Input() childData: { id: Number, name: any, key: any, message: any, protocol: any };
```
praveen-garg: protocol should be string not any

praveen-garg: id > port

src/app/modal/modal.component.ts
```diff
@@ -50,7 +50,11 @@ export class ModalComponent implements OnInit {
       this.toggleModal(false);
     }
     if (this.childData.key === 'shutdownService') {
-      this.shutdownService.emit(this.childData.id);
+      const obj = {
+        id: this.childData.id,
+        protocol: this.childData.protocol
```
praveen-garg: we should send `<SCHEME> (protocol) :// <HOST IP>: <PORT>`

src/app/services-health/services-health.component.ts
```diff
@@ -66,21 +67,22 @@ export class ServicesHealthComponent implements OnInit {
...
-      message: 'Do you really want to shut down ' + name + ' service?'
+      message: 'Do you really want to shut down ' + name + ' service',
+      protocol: protocol
     };
     // call child component method to toggle modal
     this.child.toggleModal(true);
   }
 
-  shutdownService(port) {
+  shutdownService(obj) {
```
praveen-garg: obj > svcInfo

src/app/services/services-health.service.ts
```diff
@@ -41,9 +41,13 @@ export class ServicesHealthService {
...
    *  POST  | /foglamp/service/shutdown
    */
-  shutDownService(port) {
+  shutDownService(obj) {
+    const port = obj.id;
+    const protocol = obj.protocol;
     const headers = new HttpHeaders().set(InterceptorSkipHeader, '');
-    const url = new URL(this.GET_SERVICES_URL);
+    const baseUrl = this.GET_SERVICES_URL;
+    const urlWithServiceProtocol = baseUrl.replace(/^https?/i, protocol);
```
praveen-garg: what API run on http and msvc run on https

src/app/services/services-health.service.ts
```diff
@@ -41,9 +41,13 @@ export class ServicesHealthService {
...
+    const url = new URL(urlWithServiceProtocol);
```
praveen-garg: urlWithServiceProtocol > serviceUrl

src/app/modal/modal.component.ts
monika-Dianomic: It can be; but here, `childData` is common for every modal and almost modals use it as id. Only 'Service Health' page use it as port.

PR #125, opened 2018-05-07
Added refresh icon on Assert reading and configuration component

mshariq-nerd:
## What do these changes do?
Refresh icon to load new data on Assert reading and configuration component

No

## Related issue number
None

## Checklist

### DIFF
--- a/src/app/asset-readings/assets/assets.component.html
+++ b/src/app/asset-readings/assets/assets.component.html
@@ -48,7 +48,11 @@
   <div class="card">
     <header>
       <div id="card-title" class="card-content">
-        <h5 class="title is-5">Asset Readings</h5>
+        <h5 class="title is-5">Asset Readings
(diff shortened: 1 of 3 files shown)


src/app/asset-readings/assets/assets.component.html
```diff
@@ -48,16 +48,20 @@ <h5 class="title is-5"> Assets
   <div class="card">
     <header>
       <div id="card-title" class="card-content">
-        <h5 class="title is-5">Asset Readings</h5>
+        <h5 class="title is-5">Asset Readings
+          <button (click)="getAsset(); setAssetCode(selectedAsset)" class="button is-small">
```
praveen-garg: is this a chained call? what if selectedAsset is not inside latest getAsset result? will it show `Select` or get blank?

mshariq-nerd: yes It is kind of chained call and  if `selectedAsset ` is not there then it will show `Select`.

PR #146, opened 2018-05-24
1.2.0

praveen-garg:
v1.2.0 release merge to master!


praveen-garg:
Looks like this is because develop never got v1.1.1 in, after its release!

PR #154, opened 2018-06-01
Updated angular version to v5

mshariq-nerd:
1. Angular v5.2.11
2. Angular cli v1.7.4
3. Typescript v.2.6.2
4. Removed ng-mask library because of the buggy nature on Angular 5. Added `input-mask.directive.ts` to fix the mask issue


src/app/directives/input-mask.directive.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  private state: IState;
+
+  @Input() mask: any;
+  @Output() ngModelChange = new EventEmitter();
+
+  /**
+   *
+   * @param element
+   * @param model
```
monika-Dianomic: extra param `model`, there is no param named as model.

src/app/directives/input-mask.directive.ts
```diff
@@ -0,0 +1,218 @@
...
+  'ENTER': 13
+};
+
+
+interface IState {
+  value: string;
+}
+
+@Directive({
+  selector: '[mask]'
```
monika-Dianomic: It should have prefix "app" as per `lint` check

mshariq-nerd: @monika-Dianomic  IMHO we need to keep it as it is because this is not a app-level thing like a component. Directive mainly applied to layout fields used for validation.  `[mask], [validateEqual]` makes more sense then `[appMask], [appValidateEqual]`. So we need to bypass it at lint side. 
cc/ @praveen-garg

monika-Dianomic: Ok, that's fine.

tslint.json
```diff
@@ -116,9 +116,7 @@
     ],
     "directive-selector": [
       true,
-      "attribute",
-      "app",
-      "camelCase"
+      "attribute"
```
praveen-garg: I think we should keep `camelCase` too

PR #160, opened 2018-06-08
Fix syslog print order issue

mshariq-nerd:
1. Fixed syslog print order issue


src/main.ts
```diff
@@ -6,6 +6,9 @@ import { environment } from './environments/environment';
 
 if (environment.production) {
   enableProdMode();
+  if (window) {
+    window.console.log = function () { };
+  }
```
praveen-garg: move to another PR please!

PR #165, opened 2018-06-12
Fixed validation check on Add service

mshariq-nerd:
### Fixed Issues
- [x] handle entries with space (validation)
- [x] add slight async delay to refresh services list, to show newly added.
- [x] add e2e test for Add service button
- [x] change-log feature addition


src/app/components/core/services-health/services-health.component.ts
```diff
@@ -129,5 +129,4 @@ export class ServicesHealthComponent implements OnInit {
   onNotify() {
     this.getServiceData();
   }
-
```
praveen-garg: this file check-in not needed!

PR #180, opened 2018-06-20
App login handling based on authenticationOptional value return by ping service

mshariq-nerd:
## What do these changes do?
App login handling based on authenticationOptional value return by ping service.

##### mandatory
a) On fresh startup if no service url configured then app should redirects to settings page.
b) After successful url configuration app should redirects to _login page_ 
c) On navigation bar user profile dropdown menu should be there.
d) Check allowPing and authentication mandatory case
e) Check admin and normal user login scenarios.
 

##### Optional
a) On fresh startup if no service url configured then app should redirects to settings page.
b)  After successful url configuration app should redirects on _dashboard page_
c) Only foglamp shutdown button should be visible in navigation bar ?
d)  All side menu options should work?

Yes

## Checklist

### DIFF
--- a/e2e/environment.ts
+++ b/e2e/environment.ts
@@ -1,4 +1,4 @@
 export const environment = {
-  HOST: '192.168.1.12',
+  HOST: '10.2.5.12',
   SERVICE_PORT: '8081'
 };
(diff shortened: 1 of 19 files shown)


praveen-garg:
closing in favor of #185

PR #185, opened 2018-06-25
App landing page handling on authenticationOptional return by ping request

mshariq-nerd:
## What do these changes do?
App login handling based on authenticationOptional value return by ping service.

##### mandatory
a) On fresh startup if no service url configured then app should redirects to settings page.
b) After successful url configuration app should redirects to _login page_ 
c) On navigation bar user profile dropdown menu should be there.
d) Check allowPing and authentication mandatory case
e) Check admin and normal user login scenarios.
 

##### Optional
a) On fresh startup if no service url configured then app should redirects to settings page.
b)  After successful url configuration app should redirects on _dashboard page_
c) Only foglamp shutdown button should be visible in navigation bar ?
d)  All side menu options should work?

## Checklist

### DIFF
--- a/src/app/components/layout/navbar/navbar.component.ts
+++ b/src/app/components/layout/navbar/navbar.component.ts
@@ -87,7 +87,7 @@ export class NavbarComponent implements OnInit, AfterViewInit, OnDestroy {
       }
       this.status.changeMessage(true);
       this.ping_data = data;
-      const statsTxt = 'Read:' + data['dataRead'] + '\n' + 'Sent:' + data['dataSent'] + '\n' + 'Purged:' + data['dataPurged'];
+      const statsTxt = 'Read: ' + data['dataRead'] + '\n' + 'Sent: ' + data['dataSent'] + '\n' + 'Purged: ' + data['dataPurged'];
       this.ping_info = { stats: statsTxt, is_alive: true, service_status: 'running' };
       if (JSON.parse(sessionStorage.getItem('LOGIN_SKIPPED')) === true) {
         this.isUserLoggedIn = false;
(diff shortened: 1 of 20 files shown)


src/app/components/layout/navbar/navbar.component.ts
```diff
@@ -71,42 +67,43 @@ export class NavbarComponent implements OnInit, AfterViewInit, OnDestroy {
...
-          this.ping_info = { stats: 'No data', is_alive: false, service_status: 'service down' };
-        },
-    );
+    this.servicesHealthService.pingService().then(data => {
+      if (pingManually === true) {
+        this.ngProgress.done();
+      }
+      this.status.changeMessage(true);
+      this.ping_data = data;
+      const statsTxt = 'Read:' + data['dataRead'] + '\n' + 'Sent:' + data['dataSent'] + '\n' + 'Purged:' + data['dataPurged'];
```
praveen-garg: add spaces after :

src/app/services/services-health.service.ts
```diff
@@ -12,17 +12,23 @@ export class ServicesHealthService {
...
+
   /**
      *  GET  | /foglamp/ping
      */
-  pingService() {
-    return this.http.get(this.GET_PING_URL).pipe(
-      timeout(this.REQUEST_TIMEOUT_INTERVAL),
-      map(response => response),
-      catchError((error: Response) => observableThrowError(error)));
+  pingService(): Promise<any> {
```
praveen-garg: Here we are using promise pattern instead of subscriber!

praveen-garg:
@mshariq-nerd  

1) In case of auth: mandatory + allowPing: false - we should show service status as up if 403 ; as its up just need authentication to get ping data (instead of checkmark in green use lock icon) 
screen shot 2018-06-26 at 12 11 11 am

2) if app restart (event) happens, make sure user-(profile or management) screens are routed back to home

/cc @monika-Dianomic

PR #187, opened 2018-06-26
Fixed user role based routing issue - [security fix]

mshariq-nerd:
## What do these changes do?
1) Applied guard to access user-management and profile page as per user role

## Related issue number
None 

## Checklist

### DIFF
--- a/src/app/components/core/user-management/reset-password/reset-password.component.ts
+++ b/src/app/components/core/user-management/reset-password/reset-password.component.ts
@@ -4,7 +4,6 @@ import { ActivatedRoute, Router } from '@angular/router';
 import { NgProgress } from 'ngx-progressbar';
 
 import { AlertService, UserService } from '../../../../services';
-import { SharedService } from '../../../../services/shared.service';
 
 @Component({
   selector: 'app-reset-password',
@@ -19,8 +18,7 @@ export class ResetPasswordComponent implements OnInit {
     private userService: UserService,
     public ngProgress: NgProgress,
     private activatedRoute: ActivatedRoute,
-    private router: Router,
-    private sharedService: SharedService) {
+    private router: Router) {
     // get username from url
     this.userName = this.activatedRoute.snapshot.queryParams['username'];
   }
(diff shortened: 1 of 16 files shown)


src/app/components/core/user-management/user.management.module.ts
```diff
@@ -37,7 +39,7 @@ const routes: Routes = [
     NgProgressModule,
     AlertDialogModule
   ],
-  providers: [],
+  providers: [UserGuard],
```
praveen-garg: AuthGuard is not needed here?

src/app/guards/user.guard.ts
```diff
@@ -0,0 +1,18 @@
...
+export class UserGuard implements CanActivate {
+  skip: boolean;
+  constructor(private router: Router) { }
+  canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot) {
+    const isAdmin = JSON.parse(sessionStorage.getItem('isAdmin'));
+    if (isAdmin) {
+      // logged in so return true
+      return true;
+    }
+    // not logged in so redirect to login page with the return url
```
praveen-garg: what if optional?

src/app/components/core/user-management/user.management.module.ts
mshariq-nerd: No because AuthGuard is used with other modules as well so it is inside app.module.ts

src/app/guards/user.guard.ts
mshariq-nerd: This is to check if user is admin. In all other case it will return false.

src/app/components/core/user-management/reset-password/reset-password.component.ts
```diff
@@ -17,13 +18,14 @@ export class ResetPasswordComponent implements OnInit {
   constructor(private alertService: AlertService,
     private userService: UserService,
     public ngProgress: NgProgress,
-    private activatedRoute: ActivatedRoute, private router: Router) {
+    private activatedRoute: ActivatedRoute,
+    private router: Router,
+    private sharedService: SharedService) {
```
monika-Dianomic: It seems `sharedService` is useless here.

PR #188, opened 2018-06-27
Fixed gui version mismatch issue in deploy [FOGL-1483 - gui build fails]

mshariq-nerd:


praveen-garg:
For reference linking: https://github.com/foglamp/foglamp-gui/issues/73

PR #191, opened 2018-07-02
Fixed e2e tests as per authOptional true/false

monika-Dianomic:
Fixed e2e tests as per authOptional true/false

**NOTE:** If `authenticationOptional` (returned from ping request) is true then, set variable `AUTH_OPTIONAL` to `true` in `environment.ts` otherwise set it to `false`.

### DIFF
--- a/e2e/po/app.skip.ts
+++ b/e2e/po/app.skip.ts
@@ -374,7 +374,7 @@ export class SkipLogin {
     return element(by.css('app-services-health table thead tr')).getText();
   }
 
-  httpSouthServiceStatus() {
+  coreServiceStatus() {
     browser.ignoreSynchronization = true;
     // wait
     browser.wait(this.EC.visibilityOf(element(by.css('#FogLAMP_Core td:nth-child(2) div span'))), 4000);
--- a/e2e/po/app.skip.ts
+++ b/e2e/po/app.skip.ts
@@ -10,7 +10,6 @@ export class SkipLogin {
 
   setUpInstance() {
     browser.waitForAngularEnabled(false);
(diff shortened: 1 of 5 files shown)


e2e/po/app.skip.ts
```diff
@@ -383,8 +377,8 @@ export class SkipLogin {
   httpSouthServiceStatus() {
```
mshariq-nerd: It seems core service is verified inside this method now so method name should be changed as well to coreServiceStatus

e2e/specs/app.e2e-common.ts
```diff
@@ -12,278 +13,278 @@ describe('FogLAMP gui', () => {
...
+      expect(skipLogin.getServiceStatusTitle()).toContain('Services Status');
+      expect(skipLogin.getAddServiceButton()).toContain('Add Service');
+      expect(skipLogin.getServiceStatusRefreshButton()).toEqual(true);
+
+      for (const ColumnName in ColumnsName) {
+        expect(skipLogin.getServiceHealthColNames()).toContain(ColumnsName[ColumnName]);
+      }
+      expect(skipLogin.httpSouthServiceStatus()).toContain('running');
+
+      //   Failing due to address binding.
```
mshariq-nerd: This is not failing due to address binding but because of no south service plugin configured by default in the system now. So either we need to configured the httpSouth plugin or else we need to remove this from here.

cc/ @praveen-garg

praveen-garg: remove! we will add fixtures in another ticket for services, logs, readings and all such data ..

PR #192, opened 2018-07-02
Fixed download support bundle issue in mandatory mode [FOGL-1309]

mshariq-nerd:
FOGL-1309: foglamp-gui: Allow a way to download a support bundle file through browser GET when authentication is set to mandatory (i.e. authToken header is required)

### DIFF
--- a/src/app/components/core/support/support.component.html
+++ b/src/app/components/core/support/support.component.html
@@ -22,7 +22,7 @@
           <tr *ngFor="let bundle of bundlesData">
             <td>{{ bundle }}</td>
             <td>
-              <a href="{{ supportUrl }}/{{ bundle }}" target="_blank" class="button is-text">
+              <a (click)="downloadBundle(bundle)" target="_blank" class="button is-text">
                 download
               </a>
             </td>
--- a/src/app/components/core/support/support.component.html
+++ b/src/app/components/core/support/support.component.html
@@ -22,7 +22,7 @@
           <tr *ngFor="let bundle of bundlesData">
             <td>{{ bundle }}</td>
             <td>
-              <a (click)="downloadBundle(bundle)" target="_blank" class="button is-text">
+              <a (click)="downloadBundle(bundle)" class="button is-text">
                 download
               </a>
             </td>
--- a/src/app/components/core/support/support.component.ts
+++ b/src/app/components/core/support/support.component.ts
@@ -1,3 +1,4 @@
+import { HttpClient } from '@angular/common/http';
 import { Component, OnInit } from '@angular/core';
 import { NgProgress } from 'ngx-progressbar';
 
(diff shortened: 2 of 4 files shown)


src/app/components/core/support/support.component.html
```diff
@@ -22,7 +22,7 @@
           <tr *ngFor="let bundle of bundlesData">
             <td>{{ bundle }}</td>
             <td>
-              <a href="{{ supportUrl }}/{{ bundle }}" target="_blank" class="button is-text">
+              <a (click)="downloadBundle(bundle)" target="_blank" class="button is-text">
```
praveen-garg: target="_blank"

src/app/components/core/support/support.component.ts
```diff
@@ -10,10 +11,10 @@ import { AlertService, SupportService } from '../../../services';
...
 export class SupportComponent implements OnInit {
   public bundlesData = [];
-  public supportUrl = this.supportBundleService.SUPPORT_BUNDLE_URL;
-
-  constructor(private supportBundleService: SupportService, public ngProgress: NgProgress, private alertService: AlertService) { }
-
+  constructor(private supportBundleService: SupportService,
+    public ngProgress: NgProgress,
+    private alertService: AlertService,
+    private http: HttpClient) { }
```
praveen-garg: used?

src/app/services/support.service.ts
```diff
@@ -28,4 +28,11 @@ export class SupportService {
       map(response => response),
       catchError((error: Response) => observableThrowError(error)));
   }
+
+  public async downloadSupportBundle(bundle): Promise<Blob> {
+    const file = await this.http.get<Blob>(
+      this.SUPPORT_BUNDLE_URL + '/' + bundle,
+      { responseType: 'blob' as 'json' }).toPromise();
```
praveen-garg: as json? please add docstring

praveen-garg: is http interceptor sending the token here?

src/app/components/core/support/support.component.ts
```diff
@@ -55,5 +56,17 @@ export class SupportComponent implements OnInit {
...
+
+  public async downloadBundle(bundle): Promise<void> {
+    const blob = await this.supportBundleService.downloadSupportBundle(bundle);
+    const url = window.URL.createObjectURL(blob);
+    // create a custom anchor tag
+    const a = document.createElement('a');
+    document.body.appendChild(a);
+    a.href = url;
+    a.download = bundle;
+    a.click();
```
praveen-garg: put `document.body.appendChild(a);` just above `a.click();` after assigining href and download attr/prop values

src/app/services/support.service.ts
mshariq-nerd: Type script throws an error `Types of property 'responseType' are incompatible.
Type '"blob"' is not assignable to type '"json"` if `blob as json` not used.


Yes all request have token supplied by  http interceptor

PR #193, opened 2018-07-02
Fixed system logout issue on browser hard refresh in mandatory mode

mshariq-nerd:
Fixed system logout issue on browser refresh in mandatory mode


src/app/services/http.request.interceptor.ts
```diff
@@ -35,6 +35,7 @@ export class HttpsRequestInterceptor implements HttpInterceptor {
             this.router.navigate(['/login']);
             sessionStorage.clear();
           } else if (err.status === 403) {
+            this.router.navigate(['/login']);
             sessionStorage.clear();
```
praveen-garg: clear before navigate!

PR #205, opened 2018-07-17
Restricted input fields to enter valid input type in Configuration [part of FOGL-1433]

monika-Dianomic:
- Dropdown appears for `boolean` type field
- For `numeric` type of field, `alphabetic` value should not be entered.
- For `string` type of field, any type of value can be entered.

**Note:** ['IPv4', 'IPv6', 'X509 certificate', 'password', 'JSON'] are considered as string for now


src/app/components/core/configuration-manager/configuration-manager.component.html
```diff
@@ -41,21 +41,34 @@
...
-                    (input)="onTextChange(category.key.trim() + '-' + v.key.trim())" />
+                  <div class="control is-fullwidth">
+                    <form name="form" id="ngForm" #f="ngForm" novalidate>
+                      <select *ngIf="v.value.type == 'boolean'" id="{{category.key.trim() + '-' + v.key.trim() | lowercase }}" (input)="onTextChange(category.key.trim() + '-' + v.key.trim())"
+                        (focus)="oldVal=v.value.value" class="select is-fullwidth is-small" [(ngModel)]="v.value.value" name="categoryValue"
+                        #categoryValue="ngModel" (ngModelChange)="onModelChange(oldVal);">
+                        <option value="true">true</option>
+                        <option value="false">false</option>
+                      </select>
+                      <input *ngIf="v.value.type == 'string' || v.value.type == 'IPv4' || v.value.type == 'IPv6' || v.value.type == 'X509 certificate' || v.value.type == 'password' || v.value.type == 'JSON'"
```
praveen-garg: compare after converting to  lower/upper case

Can we simply check ` UPPERCASE of v.value.type in [STRING, IPV4, .... JSON]`

src/app/components/core/configuration-manager/configuration-manager.component.html
```diff
@@ -41,21 +41,34 @@
...
-                  <input *ngIf='isObject(v?.value?.value)' id="{{category.key.trim() + '-' + v.key.trim() | lowercase }}" class="input is-small" type="text"
-                    [value]="v?.value?.value != undefined ? (JSON.stringify(v?.value?.value)) : '' " [textContent]="v?.value?.value != undefined ? (JSON.stringify(v?.value?.value)) : '' "
-                    (input)="onTextChange(category.key.trim() + '-' + v.key.trim())" />
-
-                  <input *ngIf='!isObject(v?.value?.value)' id="{{category.key.trim() + '-' + v.key.trim() | lowercase }}" class="input is-small" type="text"
-                    [value]='v?.value?.value != undefined ? v?.value?.value : "" ' [textContent]='v?.value?.value != undefined ? v?.value?.value : "" '
-                    (input)="onTextChange(category.key.trim() + '-' + v.key.trim())" />
+                  <div class="control is-fullwidth">
+                    <form name="form" id="ngForm" #f="ngForm" novalidate>
+                      <select *ngIf="(v.value.type | uppercase) == 'BOOLEAN'" id="{{category.key.trim() + '-' + v.key.trim() | lowercase }}" (input)="onTextChange(category.key.trim() + '-' + v.key.trim())"
```
praveen-garg: it looks ugly! 

Please use `<div class="select" *ngIf="(v.value.type | uppercase) == 'BOOLEAN'">` and see why Purge -> retainUnsent is not binding! Looks it has value True/ False while expected model binding is true/false!

PR #210, opened 2018-07-23
don't parse JSON; just send as is; we already have stringify

praveen-garg:

### DIFF
--- a/src/app/services/configuration.service.ts
+++ b/src/app/services/configuration.service.ts
@@ -37,7 +37,7 @@ export class ConfigurationService {
         const categoryName = encodeURIComponent(category_name);
         let body = JSON.stringify({ 'value': value });
         if (type.toUpperCase() === 'JSON') {
-            body = JSON.stringify({ 'value': JSON.parse(value) });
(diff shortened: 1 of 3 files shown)


src/app/services/configuration.service.ts
```diff
@@ -37,7 +37,7 @@ export class ConfigurationService {
         const categoryName = encodeURIComponent(category_name);
         let body = JSON.stringify({ 'value': value });
         if (type.toUpperCase() === 'JSON') {
-            body = JSON.stringify({ 'value': JSON.parse(value) });
+            body = JSON.stringify({ 'value': value });
```
praveen-garg: @monika-Dianomic @mshariq-nerd 

Once this is in,
https://github.com/foglamp/FogLAMP/pull/953#discussion_r204904754
1) we shall need `app/components/core/configuration-manager/add-config-item/add-config-item.component.html` to have textarea as well for type JSON.

praveen-garg: and this will go to 1433-main! not the develop as f953 is for f1432.

praveen-garg:
Not required!

PR #224, opened 2018-08-06
Renamed Services Health to Services and fixed e2e tests

monika-Dianomic:


e2e/environment.ts
```diff
@@ -1,5 +1,5 @@
 export const environment = {
-  HOST: '192.168.1.197',
+  HOST: '10.2.5.20',
```
praveen-garg: this should not be needed, even we shall need a script to find and replace this when run on CI

PR #228, opened 2018-08-10
GUI header: fixed for mobile view

monika-Dianomic:
**Fixed header for better mobile view.**

Due to long text of `host name`, the whole text not appearing on mobile view.

bad-view

Updated view for Mobile:

fixed-view

fixed-view1


monika-Dianomic:
@billehunt  There are three views in Header: `mobile`, `tablet` and `desktop`. I have changed only for `mobile view`. For tablet and desktop, view will be the same as it was.

billehunt:
Hmm. Let's do this: for movie only: leave the full host/service in the drop down, and display the first 10 chars of the host name, slash, first 10 chars of the service name on the header.

PR #234, opened 2018-08-13
removed a) add, delete schedule options b) timestamp format info tooltip

praveen-garg:


monika-Dianomic:
@billehunt, changes in `side-menu.component.html` are reverted back. Please review and merge.

PR #236, opened 2018-08-14
Schedule update fix

praveen-garg:


src/app/components/core/scheduler/list-schedules/list-schedules.component.ts
```diff
@@ -40,6 +45,10 @@ export class ListSchedulesComponent implements OnInit {
   constructor(private schedulesService: SchedulesService, private alertService: AlertService, public ngProgress: NgProgress) { }
 
   ngOnInit() {
+
+    this.getScheduleType();
+    this.getSchedulesProcesses();
```
praveen-garg: later these calls will be made on demand only. or on edit modal only for now. method def for these should be in some common schedule module.

PR #254, opened 2018-08-25
[FOGL-1640] filter schedule to exclude south and north

praveen-garg:
- [x] Remove South and North schedules 
- [x] Refresh Tasks pane with every ping
-------
- [x] remove dummy view-config component (it was for mockup only)


src/app/components/core/scheduler/list-schedules/list-schedules.component.ts
```diff
@@ -16,17 +18,18 @@ enum weekDays {
...
   templateUrl: './list-schedules.component.html',
   styleUrls: ['./list-schedules.component.css']
 })
+
 export class ListSchedulesComponent implements OnInit {
   public scheduleData = [];
   public scheduleProcess = [];
   public scheduleType = [];
   public days = [];
-  public scheduler_name: string;
```
praveen-garg: @mshariq-nerd is this being used for enable/disable or edit/update? 
/cc @monika-Dianomic

src/app/components/core/scheduler/list-schedules/list-schedules.component.ts
```diff
@@ -60,6 +64,73 @@ export class ListSchedulesComponent implements OnInit {
...
+            s.children.forEach( el => {
+              sn.push(el.key);
+            });
+          });
+          const north = allCats.filter(el => el.key.toUpperCase() === 'NORTH');
+          north.forEach(n => {
+            n.children.forEach( el => {
+              sn.push(el.key);
+            });
+          });
```
praveen-garg: I would love to move this to another private `function filterCategories(categories)` where `categories = data['categories']`

praveen-garg: Done

PR #261, opened 2018-08-28
Remove the extra Save buttons in the  config pages for South and North

praveen-garg:


praveen-garg:
closing in favor of #264

PR #267, opened 2018-08-30
FOGL-1757: Asset reading graphs: Show summary labels sorted by name and use single api call for an asset's summary

mshariq-nerd:
Depends on: https://github.com/foglamp/FogLAMP/pull/1056

1. Single api call for an asset's summary
2. Summary labels sorted by name
3. Asset summary refresh depends with every ping (as the assets graph)
4. Added time filter for summary and asset readings 
5. Applied limit on summary


src/app/components/core/asset-readings/assets/assets.component.ts
```diff
@@ -38,6 +39,7 @@ export class AssetsComponent implements OnInit, OnDestroy {
       subscribe(
         (data: any[]) => {
           this.assets = data;
+          this.assets = orderBy(this.assets, ['assetCode'], ['asc']);
```
praveen-garg: what if `['assetCode', asc']`

PR #274, opened 2018-08-31
FOGL-1765 : Config Page improvements

mshariq-nerd:
1) All true/false config items changed to checkbox (instead of dropdown true/false)


billehunt:
looks good

PR #275, opened 2018-08-31
[FOGL-1775]  close modal on save

praveen-garg:


billehunt:
This is resolved in latest develop.

PR #303, opened 2018-09-13
Changed version to 1.4.0

mshariq-nerd:


praveen-garg:
Branch name was not changed. 

This change was proposed to let help identify the build and debian packages name like `foglamp-gui-1.4.0rc.tar.gz` | `foglamp-gui-1.4.0rc.deb`

praveen-garg:
Reopened again, version fix as `1.4.0` instead of calling `1.4.0-dev` or `1.4.0rc`.

PR #320, opened 2018-09-27
FOGL-1838: GUI (SOUTH wizard) creating items at each stage prevents effective use of back and precludes cancelling

mshariq-nerd:


src/app/components/core/south/add-service-wizard/add-service-wizard.component.ts
```diff
@@ -118,52 +117,44 @@ export class AddServiceWizardComponent implements OnInit {
...
       case 2:
+        this.viewConfigItemComponent.callFromWizard();
         document.getElementById('vci-proxy').click();
         if (this.viewConfigItemComponent !== undefined && !this.viewConfigItemComponent.isValidForm) {
           return false;
         }
-        nxtButton.textContent = 'Done';
-        previousButton.textContent = 'Previous';
-
+        this.addService(this.payload, nxtButton, previousButton);
```
praveen-garg: this should happen only at 3rd step, after Enabled or not decision step!

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
```diff
@@ -117,4 +125,24 @@ export class ViewConfigItemComponent implements OnInit, OnChanges {
...
+    if (configVal.value !== undefined && configVal.value !== '') {
+      return configVal.value;
+    } else {
+      return configVal.default;
+    }
+  }
+
+  /**
+   * Method to set isWizardCall = true if called from
+   * Add south wizard.
```
praveen-garg: or north

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
```diff
@@ -14,10 +14,12 @@ import ConfigTypeValidation from '../configuration-type-validation';
 export class ViewConfigItemComponent implements OnInit, OnChanges {
   @Input() categoryConfigurationData: any;
   @Input() useProxy: 'false';
+  @Output() onConfigChanged: EventEmitter<any> = new EventEmitter<any>();
 
   public categoryConfiguration;
   public configItems = [];
   public isValidForm: boolean;
+  public isWizardCall: boolean;
```
praveen-garg: I think we should set the default value as `false`

PR #323, opened 2018-10-04
Added functionality to get time series graph for Assets reading [FOGL-1524]

monika-Dianomic:


src/app/components/core/asset-readings/series-graph/series-graph.component.html
```diff
@@ -0,0 +1,47 @@
...
+              <span>
+                <b>{{asset.name}} &nbsp;</b>
+              </span>
+              <span class="tag is-light"> Avg </span>
+              <span class="tag is-info" [ngStyle]="{'background-color': '#85C1E9'}">{{roundTo(reading.average, 5)}}
+              </span> &nbsp;
+              <span class="tag is-light"> Min </span>
+              <span class="tag is-info" [ngStyle]="{'background-color': '#85C1E9'}">{{roundTo(reading.min, 5)}}</span>
+              &nbsp;
+              <span class="tag is-light"> Max </span>
```
praveen-garg: I don;t think we need avg,min max here

src/app/components/core/asset-readings/series-graph/series-graph.component.ts
```diff
@@ -0,0 +1,289 @@
...
+  public assetChartType: string;
+  public assetReadingValues: any;
+  public assetChartOptions: any;
+  public showGraph = true;
+  public readings: any;
+  public assetReadingSeries = [];
+  public assetReadingSummary = [];
+  public graphRefreshInterval = POLLING_INTERVAL;
+  public optedGroup = 'seconds';
+  public readKeyColorLabel = [];
```
praveen-garg: are these all variables required?

src/app/components/core/asset-readings/series-graph/series-graph.component.ts
```diff
@@ -0,0 +1,289 @@
...
+
+  private isAlive: boolean;
+  @Output() notify: EventEmitter<any> = new EventEmitter<any>();
+  @ViewChild('assetChart') assetChart: Chart;
+
+  constructor(private assetService: AssetsService, private alertService: AlertService,
+    private ping: PingService) {
+    this.assetChartType = 'line';
+    this.assetReadingValues = [];
```
praveen-garg: assetReadings?

src/app/components/core/asset-readings/series-graph/series-graph.component.ts
```diff
@@ -0,0 +1,289 @@
...
+  public getSeriesGraph(assetCode, autoRefresh = true) {
+    this.notify.emit(false);
+    if (this.graphRefreshInterval === -1) {
+      this.isAlive = false;
+    } else {
+      this.isAlive = true;
+    }
+    this.assetCode = assetCode;
+    if (autoRefresh === false) {
+      this.plotReadingsGraph(assetCode);
```
praveen-garg: plotReadingsGraph?

src/app/components/core/asset-readings/series-graph/series-graph.component.ts
```diff
@@ -0,0 +1,289 @@
...
+      });
+      const timestamps = assetChartRecord.reverse().map(t => t.timestamp);
+      timestamps.forEach(timestamp => {
+        assetTimeLabels.push(datePipe.transform(timestamp, 'HH:mm:ss'));
+      });
+    }
+    this.statsAssetSeriesGraph(assetTimeLabels, assetReading);
+  }
+
+  getColorCode(readKey, cnt, fill) {
```
praveen-garg: looks it will be util function

src/app/services/assets.service.ts
```diff
@@ -58,11 +58,17 @@ export class AssetsService {
...
+    return this.http.get(this.GET_ASSET + '/' + encodeURIComponent(assetCode)).pipe(
+      map(response => response),
+      catchError((error: Response) => observableThrowError(error)));
+  }
 
-  // TODO: Not in use yet
-  public getAssetAverage(assetObject: any) {
-    // TODO: time based readings average;
-    return this.http.get(this.GET_ASSET + '/' + encodeURIComponent(assetObject.assetCode) + '/' + assetObject.reading + '/series').pipe(
+  public getAssetAverage(assetCode: string, reading: string, group) {
```
praveen-garg: average?

mshariq-nerd:
In case of empty value in time field. `foglamp/asset/system%2FcpuUsage%2Fall/%iowait/series?group=seconds&seconds=null` throwing error `400 Time must be a positive integer`

mshariq-nerd:
Still seeing this issue 
for blank value  in input field

> foglamp/asset/system%2FcpuUsage%2Fall/%iowait/series?group=seconds&seconds=null

 throwing error 400 Time must be a positive integer

mshariq-nerd:
@billehunt This PR is ready for final review. You can test it with FogLAMP `develop` branch

src/app/components/core/asset-readings/series-graph/series-graph.component.html
```diff
@@ -0,0 +1,47 @@
...
+      </div>
+    </header>
+    <section class="modal-card-body">
+      <div *ngIf="showSpinner" class="app-loading">
+        &nbsp;
+        <svg class="spinner" viewBox="0 0 100 100">
+          <circle class="path" cx="50%" cy="50%" r="10" fill="none" stroke-width="2" stroke-miterlimit="10" />
+        </svg>
+      </div>
+      <div *ngIf='!showSpinner && showGraph' class="columns is-mobile">
```
praveen-garg: we should remove `showGraph` from here? 
As this is being set on `ReadingsValidator.validate(data); ` so if data for grpah is not numeric, allow user to select asset/sensor to fetch  series  for newly selected item

src/app/utils.ts
```diff
@@ -80,6 +80,25 @@ export default class Utils {
     return totalSec < 0 || totalSec >= 86400;
   }
 
+  public static getColorCode(readKey, cnt) {
```
praveen-garg: as we have it in Utils, can we remove the copy for readings graph and use this itself.

billehunt:
I'm afraid this functionality is not useful. Users don't need a graph of min/max/average over time.

Please don't add any graphs without having me in the loop.

praveen-garg:
Closing as not required!

billehunt:
Actually, I think I jumped the gun with my last comment. My assumption was that this was for the ongoing min, max and average (as the graph above in this PR shows). But now that I'm running the code, I see you're actually looking at min/max/avg for each of the time slices displayed. Which is really good, in fact, way better than what I was thinking (just an avg for each data point).

Please re-open and let's discuss next steps with this. I think you're on a good path. Maybe we shouldn't be showing 3 lines, but rather something like this: https://www.researchgate.net/profile/Ian_C_Dunican/publication/311517204/figure/fig4/AS:667673046294537@1536197067559/Between-conditions-comparison-of-max-distance-m-run-displaying-the-group-average.png

Sorry about that.

PR #326, opened 2018-10-10
FOGL-1948: Repeat time interval not converting correctly in "hh:mm:ss" formate on North Add Task wizard

mshariq-nerd:


mshariq-nerd:
wrong branch name  issue.

PR #327, opened 2018-10-11
FOGL-1948: Repeat time interval not converting correctly in "hh:mm:ss" formate on North Add Task wizard

mshariq-nerd:


src/app/components/core/north/north-task-modal/north-task-modal.component.ts
```diff
@@ -125,7 +126,7 @@ export class NorthTaskModalComponent implements OnInit, OnChanges {
           /** request completed */
           this.ngProgress.done();
           this.alertService.success('Schedule updated successfully.');
-          this.notify.emit();
+          // this.notify.emit();
```
praveen-garg: we want to remove this? what this emit was doing earlier?

mshariq-nerd: No, It was a mistake. restored

PR #332, opened 2018-10-16
Upload script with formData in add service wizard [part of FOGL-1814]

monika-Dianomic:
Test with https://github.com/foglamp/FogLAMP/pull/1186


praveen-garg:
let;s keep the branch!

PR #334, opened 2018-10-22
Added separate dropdown for group by in time series graph

monika-Dianomic:


src/app/components/core/asset-readings/series-graph/series-graph.component.html
```diff
@@ -9,27 +9,49 @@
...
-            <app-number-input-debounce placeholder="time (default 10)" (value)="setTimeValue($event)" [val]="timeValue"></app-number-input-debounce>
-          </p>
-          <p class="control">
-            <span class="select is-small">
-              <select #selectedTime (change)="setGroup(selectedTime.value)">
-                <option value="minutes" [selected]="optedGroup == 'minutes'">minutes</option>
-                <option value="hours" [selected]="optedGroup == 'hours'">hours</option>
-                <option value="seconds" [selected]="optedGroup == 'seconds'">seconds</option>
+      <div class="columns">
+        <div class="column is-3">
```
praveen-garg: why this is is-3, I think this should be normal column

src/app/components/core/asset-readings/series-graph/series-graph.component.html
```diff
@@ -9,27 +9,49 @@
...
+          <div class="field">
```
praveen-garg: apply float right?

PR #336, opened 2018-10-24
FOGL-1887: Showing service status on mouse hover on the green/yellow/red traffic light in the status bar at the top.

mshariq-nerd:


billehunt:
Looks great. Please go forward with status icon, service name and service status in the dropdown.

billehunt:
In the current FOGL-1887 branch, hovering shows the dropdown box, which looks great. But moving the cursor away leaves the hostname/fogname as a link (blue text).

Also, clicking on that blued link locks the dropdown box in place, so moving away continues to show it. (and then you can't make it go away, by re-clicking on the link or clicking elsewhere on the page)

billehunt:
also, if a south service is disabled, the hover dropdown text shows that service as red/down. Should show it as gray/disabled (same color as the disabled status button in the south page.

PR #337, opened 2018-10-26
Delete north instance task [FOGL-1928 B PART]

mshariq-nerd:
This PR is on top of https://github.com/foglamp/foglamp-gui/pull/335. So delete service or task both can be test here. Test against [FOGL-1928-task-api](https://github.com/foglamp/FogLAMP/pull/1201)

screen shot 2018-11-01 at 2 38 40 pm
screen shot 2018-11-01 at 2 38 23 pm

### DIFF
--- a/src/app/components/common/alert-dialog/alert-dialog.component.html
+++ b/src/app/components/common/alert-dialog/alert-dialog.component.html
@@ -9,10 +9,15 @@
     <section *ngIf="shutDownServiceData" class="modal-card-body">
       {{shutDownServiceData.message}}?
     </section>
+    <section *ngIf="deleteTaskData" class="modal-card-body">
+      {{deleteTaskData.message}}?
+    </section>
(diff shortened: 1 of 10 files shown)


billehunt:
Please
* remove the left "Delete Instance" text
* change the link text to say "Delete Service" (and leave it right justified)
* change the link text to be a button that's white like the cancel button
* make sure there's a confirmation box with text "Are you sure you want to delete this service?" with a default white "Cancel" button and a red "Delete" button.

mshariq-nerd:
@billehunt  Changes done as per your feedback but for the last point we already have the  confirmation box with text **Warning: Deleting this service can not be un-done. Continue?** . that text is given in FOGL-1928.  

screen shot 2018-10-31 at 5 20 08 pm

Please confirm we are good with the current message? or want to change it by the text given in the feedback above .

PR #342, opened 2018-11-12
Removed empty validation check on save config items [FOGL-2020]

mshariq-nerd:


src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
```diff
@@ -61,7 +61,9 @@ export class ViewConfigItemComponent implements OnInit, OnChanges {
 
   public difference(obj, bs) {
```
praveen-garg: Looks like we should have BaseConfigComponenent or some common module for such methods

PR #345, opened 2018-11-26
[FOGL:2106] Added config items save functionality to use bulk update api

mshariq-nerd:
This PR need to be tested against https://github.com/foglamp/FogLAMP/pull/1251

### DIFF
--- a/src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
+++ b/src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
@@ -1,6 +1,6 @@
 import { Component, EventEmitter, Input, OnChanges, OnInit, Output, SimpleChanges } from '@angular/core';
 import { NgForm } from '@angular/forms';
-import { differenceWith, find, sortBy } from 'lodash';
+import { differenceWith, sortBy, isEqual, isEmpty } from 'lodash';
 import { NgProgress } from 'ngx-progressbar';
 
 import { AlertService, ConfigurationService } from '../../../../services';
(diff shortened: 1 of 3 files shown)


src/app/services/configuration.service.ts
```diff
@@ -101,4 +101,14 @@ export class ConfigurationService {
...
+
+  /**
+  *  PUT  | /foglamp/category/{categoryName}
+  */
+  updateBulkConfiguration(categoryName, configItems) {
+    categoryName = encodeURIComponent(categoryName);
+    return this.http.put(this.CATEGORY_URL + '/' + categoryName, configItems).pipe(
+      map(response => response),
+      catchError((error: Response) => observableThrowError(error)));
+  }
```
monika-Dianomic: 'saveConfigItem' method can be removed from `configuration.service.ts` as it is not in use now.

src/app/components/core/south/south-service-modal/south-service-modal.component.ts
```diff
@@ -195,9 +195,20 @@ export class SouthServiceModalComponent implements OnInit, OnChanges {
    * @param changedConfig changed configuration of a selected plugin
    */
   getChangedConfig(changedConfig) {
```
monika-Dianomic: This method should not call in the case of clicking `Show/Hide Advanced Config` in order to remove the error in console `ERROR TypeError: Cannot read property 'toUpperCase' of undefined`

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
```diff
@@ -34,75 +34,63 @@ export class ViewConfigItemComponent implements OnInit, OnChanges {
...
     if (this.isWizardCall) {
-      this.onConfigChanged.emit(diff);
+      this.onConfigChanged.emit(changedConfigValues);
       return;
     }
-    diff.forEach(changedItem => {
-      this.saveConfigValue(this.categoryConfiguration.key, changedItem.key, changedItem.value, changedItem.type);
-    });
+
+    this.updateConfiguration(this.categoryConfiguration.key, changedConfigValues);
```
monika-Dianomic: there should be check, `updateConfiguration` method should not call if changedConfigValues.length == 0

src/app/components/core/south/south-service-modal/south-service-modal.component.ts
praveen-garg: @monika-Dianomic  this seems to get fixed with

```
    if (isEmpty(changedConfig)) {
      return;
    }
```

PR #358, opened 2018-12-19
FOGL-2205: Filter GUI feedback fix

mshariq-nerd:
- [x] Fixed Initial open animation of south modal 
- [ ] Adding the RMS filter shows enable option in middle of options. Enable should always be last   [FOGL-2206](https://scaledb.atlassian.net/browse/FOGL-2206)
- [ ] “enable” should instead be “enabled” as it is for the south/north service.  [FOGL-2206](https://scaledb.atlassian.net/browse/FOGL-2206)
- [x] icon to open folded app is currently a “+”. It should be a down arrow. 
- [ ] opening the “delta” filter doesn’t fit in the popup. **(Not able to reproduce)**
- [x] Added  hamburger icon on left of filter in South config to indicate the user can grab that to move it. 
- [x] When cursor hovers over that hamburger, icon change to a grab icon

### DIFF
--- a/src/app/components/core/filter/add-filter-wizard/add-filter-wizard.component.ts
+++ b/src/app/components/core/filter/add-filter-wizard/add-filter-wizard.component.ts
@@ -105,7 +105,8 @@ export class AddFilterWizardComponent implements OnInit {
         nxtButton.textContent = 'Next';
         previousButton.textContent = 'Previous';
 
-         // To verify if filter with given name already exist
+         // To verify if category (or filter itself) with this name already exists
+         // hence filter can not be created with that name
          const isFilterExist = this.categories.some(item => {
           return formValues['name'].trim() === item.key;
         });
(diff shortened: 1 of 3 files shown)


src/app/components/core/filter/add-filter-wizard/add-filter-wizard.component.ts
```diff
@@ -102,6 +105,15 @@ export class AddFilterWizardComponent implements OnInit {
         nxtButton.textContent = 'Next';
         previousButton.textContent = 'Previous';
 
+         // To verify if filter with given name already exist
```
praveen-garg: // To verify if category (or filter itself) with this name already exists hence filter can not be created with that name

src/app/components/core/south/south-service-modal/south-service-modal.component.html
```diff
@@ -49,16 +49,17 @@
                 <article *ngFor="let filter of filterPipeline; let i = index" [dndDraggable]="filter" (dndStart)="onDragStart(i)"
                   class="is-light accordion" [attr.id]="i">
 
-                  <div class="accordion-header" (click)="activeAccordion(i, filter)">
-                    <p>{{filter}}</p>
-                    <button class="is-dark toggle" aria-label="toggle"></button>
+                  <div class="accordion-header">
+                    <span><span class="switch-icon"><i class="fa fa-bars"></i><i class="fa fa-hand-rock-o"></i></span>&nbsp;{{filter}}</span>
```
praveen-garg: set opacity to 0.5 for fa-bars and fa-hand-rock-o

PR #360, opened 2018-12-21
FOGL-1921 : Add support to insert data processing plugin into (North)	

mshariq-nerd:


src/app/components/core/filter/add-filter-wizard/add-filter-wizard.component.ts
```diff
@@ -260,7 +260,10 @@ export class AddFilterWizardComponent implements OnInit {
   public getInstalledFilterPlugins() {
     this.filterService.getInstalledFilterPlugins().subscribe(
       (data: any) => {
-        this.plugins = data.plugins;
+        console.log('data.plugins', data.plugins);
```
praveen-garg: remove this please

PR #364, opened 2018-12-28
use node v10.x LTS

praveen-garg:


prerequisite-ubuntu.md
```diff
@@ -5,7 +5,7 @@
 Install nodejs (version 8)
```
monika-Dianomic: It should also change to `version 10`

PR #367, opened 2019-01-03
FOGL-2261: filter gui issues

mshariq-nerd:
- [x] when cursor moves over a filter, the cursor should change to a "move cursor" the hamburger in the filter should not change

- [ ] when the filter is moved, it should disappear from the list. So if there are two filters, rms and delta, and the user grabs rms and moves it, the rms filter should disappear from the list, so only delta is displayed, and the user has two options, to drop rms above delta, or below it. (currently, rms and delta both display, and user has three possibilities: drop above rms, between rms & delta, or below delta)

If there is only one filter, re-arranging capability should not be allowed. i.e.
- [x] hovering over a single filter should not change to the move cursor, and 
- [ ] that filter cannot be moved.

Also:
- [x] when a filter is expanded / collapsed, the "open" icon / the "close" should be centered
- [x] Fixed drag issue on filter config area [drag is applicable on header only]

### DIFF
--- a/src/app/components/core/north/north-task-modal/north-task-modal.component.css
+++ b/src/app/components/core/north/north-task-modal/north-task-modal.component.css
@@ -29,26 +29,18 @@ a { color: black; }
 
 .toggle::before{
   content:'\25BC' !important;
-  padding: 4px 3px 0 3px !important;
-  text-align: center !important;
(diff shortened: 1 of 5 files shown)


mshariq-nerd:
This will be closed in favor of #370

PR #372, opened 2019-01-10
FOGL-2269: Selecting item from plugin list adds description but make dialog resize

mshariq-nerd:
**Fixed Items**
1) Selecting item from plugin list adds description but make dialog resize issue

screen shot 2019-01-10 at 3 57 59 pm

2) Message for no or only readonly configuration items on south and north wizard (Rare case, e.g. `game` plugin has one config item only and even that is  `readonly`)

&nbsp;&nbsp; screen shot 2019-01-10 at 3 58 31 pm

### DIFF
--- a/src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
+++ b/src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
@@ -16,6 +16,7 @@ export class ViewConfigItemComponent implements OnInit, OnChanges {
   @Input() useProxy: 'false';
   @Input() useFilterProxy: 'false';
   @Input() formId: '';
+  @Input() pageId: 'page';
   @Output() onConfigChanged: EventEmitter<any> = new EventEmitter<any>();
 
   public categoryConfiguration;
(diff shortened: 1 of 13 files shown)


src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
```diff
@@ -57,6 +59,15 @@ export class ViewConfigItemComponent implements OnInit, OnChanges {
             type: el.type
           };
         });
+        // Verify if all config object have a readonly key
```
praveen-garg: // check if editable config item found, based on readonly property

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
```diff
@@ -57,6 +59,15 @@ export class ViewConfigItemComponent implements OnInit, OnChanges {
...
+        for (const el of this.categoryConfiguration.value) {
+          if (!has(el, 'readonly')) {
+            this.hasEditableConfigItems = true;
+            break;
+          } else {
+            this.hasEditableConfigItems = false;
```
praveen-garg: this should have a default value outside the for loop Or at class level

PR #376, opened 2019-01-18
Subscription call collapse with select graph call on dashboard [FOGL-2333]

monika-Dianomic:


src/app/components/core/dashboard/dashboard.component.ts
```diff
@@ -224,6 +224,9 @@ export class DashboardComponent implements OnInit, OnDestroy {
             }
           });
         });
+        if (this.isAlive === false) {
+          this.isAlive = true;
+        }
```
praveen-garg: if we have to set it to true, then if check can be avoided; as it can have only true / false and we want to set true anyway

src/app/components/core/dashboard/dashboard.component.ts
```diff
@@ -240,6 +243,7 @@ export class DashboardComponent implements OnInit, OnDestroy {
   }
 
   public checkedGraph(event) {
+    this.isAlive = false;
```
praveen-garg: why? please add a comment here; and also with isAlive declaration (purpose comment, as variable name is not self explantaory)

src/app/components/core/dashboard/dashboard.component.ts
praveen-garg: This change is not required!

src/app/components/core/dashboard/dashboard.component.ts
praveen-garg: This change is not required!

PR #377, opened 2019-01-18
Fixed undefined graph issue on Dashboard [FOGL-2333]

monika-Dianomic:


src/app/components/core/dashboard/dashboard.component.ts
```diff
@@ -72,7 +72,7 @@ export class DashboardComponent implements OnInit, OnDestroy {
       this.selectedGraphsList.push(selectedGraph);
     }
 
-    // if there is no graph selected, set default to "READINGS" and "North Readings to PI"
+    // if there is no graph selected, set default to "READINGS" and "READINGS SENT"
```
praveen-garg: its case sensitive actually `Readings Sent`

src/app/components/core/dashboard/dashboard.component.ts
```diff
@@ -101,7 +101,9 @@ export class DashboardComponent implements OnInit, OnDestroy {
         for (const graph of this.selectedGraphsList) {
           const selectedGraph = this.statistics.filter(value => value['key'] === graph.key);
```
mshariq-nerd: Can't we use `find` instead of `filter`? here because there will be only one returned value and `filter` always return array. that is why we have to use  `selectedGraph[0]`. I think we can eliminate that by using find?

PR #381, opened 2019-01-31
Fixed e2e tests

monika-Dianomic:


e2e/specs/app.e2e-common.ts
```diff
@@ -17,6 +17,8 @@ describe('FogLAMP gui', () => {
       skipLogin.navigateToHome();
       skipLogin.setUpInstance();
       isSetupInstance = true;
+      skipLogin.navToBackupRestore();
+      skipLogin.clickRequestBackup();
```
praveen-garg: I don;t think this should be here! this is  for before each! Move to test `it('Should Display Backup & Restore',` navToBackupRestore is already there

PR #382, opened 2019-02-01
Fixed graph issue for asset's multiple read keys [FOGL-2383]

monika-Dianomic:


src/app/components/core/asset-readings/readings-graph/readings-graph.component.html
```diff
@@ -10,57 +10,65 @@
...
-                <option value="28800" [selected]="optedTime == 28800">8 hours</option>
-                <option value="86400" [selected]="optedTime == 86400">1 day</option>
-              </select>
+        <div *ngIf="showSpinner" class="app-loading">
+          &nbsp;
+          <svg class="spinner" viewBox="0 0 100 100">
+            <circle class="path" cx="50%" cy="50%" r="10" fill="none" stroke-width="2" stroke-miterlimit="10" />
+          </svg>
+        </div>
+        <ng-container *ngIf='!showSpinner'>
```
praveen-garg: instead of recreating the graph, it should show spinner on top of current graph and update dataset of current graph object

src/app/components/core/asset-readings/readings-graph/readings-graph.component.css
```diff
@@ -10,4 +10,17 @@
...
+  width: auto;
+  min-height: 15vh;
+  max-height: 25vh;
+  overflow-y: scroll;
+  padding-top: 3px;
+  padding-bottom: 3px;
+}
+
+.modal-card {
+  width: 70%;
```
praveen-garg: Need to avoid the scrolling behavior within individual blocks of modal. Modal should have only one scroll bar.

PR #387, opened 2019-02-08
update develop with 1.5.0 rc

praveen-garg:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "foglamp",
-  "version": "1.4.5rc",
+  "version": "1.5.0RC",
```
ashish-jabble: Can’t we use 1.5.0 only?

praveen-garg: We will, when finally ready to tag!

PR #391, opened 2019-02-14
FOGL-2412 Added add south service and add filter e2e test

mshariq-nerd:


e2e/po/south.page.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  clickAddServiceButton() {
+    browser.ignoreSynchronization = true;
+    return element(by.id('add_south_service')).click();
+  }
+
+  addSouthService() {
+    browser.ignoreSynchronization = true;
+    browser.sleep(2000); // wait to get south service plugin from server
```
praveen-garg: can we poll and exit with some retries, 5 * 500ms

e2e/po/south.page.ts
```diff
@@ -0,0 +1,67 @@
...
+    element(by.name('type')).all(by.tagName('option'))
+      .then(options => {
+        options[0].click();  // select first plugin in select box
+      });
+    element(by.id('name')).sendKeys('demo-service'); // supply service name
```
praveen-garg: instead of keeping this in page object, can we pass from test? I would like the name to be `guiE2eService #1`

e2e/po/south.page.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  getServiceName() {
+    browser.ignoreSynchronization = true;
+    return element(by.css('.content table tr:nth-child(1) .button')).getText();
+  }
+
+  openSouthServiceModal() {
+    browser.ignoreSynchronization = true;
+    browser.wait(this.EC.visibilityOf(element(by.css('.content table tr:nth-child(1) .button'))), 2000);
```
praveen-garg: let's keep this wait time as constant on top of file as DETERMINISTIC_WAIT = 2000; // in ms
Dictionary

e2e/po/south.page.ts
```diff
@@ -0,0 +1,67 @@
...
+    return element(by.css('.content table tr:nth-child(1) .button')).click();
+  }
+
+  openFilterWizard() {
```
praveen-garg: should filter be another page object? we should be able to use that with north too

e2e/po/south.page.ts
```diff
@@ -0,0 +1,67 @@
...
+    browser.ignoreSynchronization = true;
+    return element(by.css('.add-application a')).click();
+  }
+
+  addFilter() {
+    browser.ignoreSynchronization = true;
+    browser.sleep(2000); // wait to get south service plugin from server
```
praveen-garg: wait and retries as above

e2e/po/south.page.ts
```diff
@@ -0,0 +1,67 @@
...
+    element(by.name('type')).all(by.tagName('option'))
+      .then(options => {
+        options[0].click();  // select first plugin in select box
+      });
+    element(by.id('name')).sendKeys('Test'); // supply service name
+    element(by.id('next')).click();
+    element(by.id('next')).click();
+    browser.wait(this.EC.visibilityOf(element(by.css('.accordion.card.cdk-drag'))), 3000);
```
praveen-garg: const on top? as above

src/app/components/core/south/south.component.html
```diff
@@ -3,7 +3,7 @@
     <div id="scheduled-process" class="card-header-title">
       <div class="media-content">
         <p class="title is-5">
-          South Services &nbsp;
+          South Services
```
praveen-garg: Are we okay now without `nbsp;` ?

e2e/specs/app.e2e-common.ts
```diff
@@ -35,16 +39,36 @@ describe('FogLAMP gui', () => {
...
+    it('Should display south service page', () => {
+      southPage.navToSouthPage();
+      expect(southPage.getSouthPageTitle()).toEqual('South Services');
+    });
+
+    it('Should display added south service', () => {
+      southPage.navToSouthPage();
+      southPage.clickAddServiceButton();
+      southPage.addSouthService();
+      expect(southPage.getServiceName()).toEqual('demo-service');
```
monika-Dianomic: I believe, created service should also be deleted in the test.

src/app/components/core/south/south.component.html
mshariq-nerd: yes, no issues. It was there to put gap between refresh button and title but now we don't have refresh button because of auto refresh functionality.

praveen-garg:
Please rebase with the develop

e2e/po/app.filters.ts
```diff
@@ -0,0 +1,66 @@
+import { browser, by, element, ExpectedConditions, promise } from 'protractor';
+
+export class Filters {
+  DETERMINISTIC_WAIT = 3000; // in milliseconds
+  retryAttempts = 0;
```
praveen-garg: why this is 0? this should be set to MAX here and we should try in test until exhausted. e.g. retryAttempts-- and retryAttempts > 0

e2e/po/app.filters.ts
```diff
@@ -0,0 +1,66 @@
...
+      // click next button twice
+      element(by.id('next')).click();
+      element(by.id('next')).click();
+    })
+      .catch((error) => {
+        if (this.retryAttempts <= 5) {
+          console.log('Retrying load filter plugin.');
+          this.addFilter(filterName);
+        } else {
+          console.log('Rejecting the promise after 5 attempts.');
```
praveen-garg: don;t hard code 5, instead use class level variable value. We would need to have it pass through args in longer term

e2e/specs/app.e2e-south.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  it('Should display added south service', () => {
+    southPage.navToSouthPage();
+    southPage.clickAddServiceButton();
+    southPage.addSouthService('guiE2eService #1'); // pass south service name
+    expect(southPage.getServiceName()).toEqual('guiE2eService #1');
+  });
+
+  it('Should display added filter in south service', () => {
+    southPage.navToSouthPage();
```
praveen-garg: is this  southPage.navToSouthPage(); required?

e2e/specs/app.e2e-south.ts
```diff
@@ -0,0 +1,35 @@
...
+    southPage.openSouthServiceModal();
+    filters.openFilterWizard();
+    filters.addFilter('guiE2eFilter #1');
+    expect(filters.getAddedFilterName()).toEqual(' guiE2eFilter #1');
+    southPage.closeSouthServiceModal();
+  });
+
+  it('Should display asset count on south service', () => {
+    southPage.navToSouthPage();
```
praveen-garg: is this  southPage.navToSouthPage(); required?

PR #405, opened 2019-03-11
FOGL-2576: Certificate upload feature is not working

mshariq-nerd:


praveen-garg:
> Alert (error) message content is not being displayed; if we choose x.key and y.cert (i.e. name does not match and backend raise 400 bad request)

As FOGL-2581 logged for this and this need to be fixed at many places. Approving this change.

PR #414, opened 2019-04-03
FOGL-2677: removed extra call for schedules filtering and use schedule processName

mshariq-nerd:


src/app/components/core/scheduler/list-schedules/list-schedules.component.ts
```diff
@@ -167,9 +102,27 @@ export class ListSchedulesComponent implements OnInit {
     this.ngProgress.start();
     this.schedulesService.getSchedules().
       subscribe(
-        (data) => {
-          // To filter
-          this.filterSouthAndNorth(data['schedules']);
+        (data: any) => {
+          data.schedules.forEach(sch => {
+            if (sch.processName !== 'south_c' && sch.processName !== 'north_c'
+              && sch.processName !== 'notification_c') {
```
praveen-garg: `if ( ! ["south_c", "north_c", "south", "north"].includes(sch.processName) ) {`

We are not using `south` for python based south plugins (services), but `north` is being used for python north plugins (tasks). so please make it ["south_c", "north_c", "north"] or better `["south_c", "north_c", "south", "north"]` even if south is not being used.

let's add `notification_c` in corresponding branch/ pull request only.

PR #428, opened 2019-04-24
Removed unused code [FOGL-2679(b)]

monika-Dianomic:


monika-Dianomic:
> deleteService is public, others are not. Any specific reason?

It doesn't have any specific reason, 'public' keyword can be removed from here.

PR #432, opened 2019-05-02
UI fixes [FOGL-2769]

monika-Dianomic:


src/app/components/layout/navbar/navbar.component.html
```diff
@@ -67,14 +67,14 @@
                 </tr>
               </table>
             </div>
-            <hr class="dropdown-divider">
+            <hr class="dropdown-divider" *ngIf="servicesRecord.length !== 0" >
```
praveen-garg: do we need this here as well?

src/app/components/layout/navbar/navbar.component.html
```diff
@@ -141,14 +141,14 @@
                     </tr>
                   </table>
                 </div>
-                <hr class="dropdown-divider">
+                <hr class="dropdown-divider" *ngIf="servicesRecord.length !== 0">
```
praveen-garg: do we need this here as well?

src/app/components/layout/navbar/navbar.component.html
```diff
@@ -192,7 +192,7 @@
                 </a>
               </div>
               <div class="dropdown-menu" id="dropdown-menu" role="menu">
-                <div class="dropdown-content">
+                <div *ngIf="servicesRecord.length !== 0" class="dropdown-content">
```
praveen-garg: do we need this here?  

Are we missing on  `<ng-container *ngIf="!showSpinner">`

src/app/components/layout/navbar/navbar.component.html
monika-Dianomic: yes, because if there is no service list below <hr>, then there is no sense to show <hr>

mshariq-nerd:
@monika-Dianomic  please clear the service record array on logout of the user. So that on login page old data should not be displayed in service dropdown (in forbidden state)

PR #433, opened 2019-05-02
FOGL-2767: Add deploy script for RHEL and centOS

mshariq-nerd:


deploy-rh
```diff
@@ -0,0 +1,239 @@
...
+echo    # (optional) move to a new line
+if [[ $REPLY =~ ^[Nn]$ ]]
+then
+    exit 0;
+fi
+
+machine_details() {
+  OS_NAME=`sudo cat /etc/*-release | grep PRETTY_NAME | cut -d= -f2-`
+  echo -e "${CINFO}Operating System: ${CRESET} ${OS_NAME} "
+  if [[ ${os} == "Darwin" ]]
```
praveen-garg: Where are we getting os value?

deploy-rh
```diff
@@ -0,0 +1,240 @@
...
+    exit 0;
+fi
+
+machine_details() {
+  os=$(uname)
+  OS_NAME=`sudo cat /etc/*-release | grep PRETTY_NAME | cut -d= -f2-`
+  echo -e "${CINFO}Operating System: ${CRESET} ${OS_NAME} "
+  if [[ ${os} == "Darwin" ]]
+  then
+      echo -e Error:"${CERR} Script ${FOGLAMP_GUI_VER} is not compatible with macOS ${CRESET}"
```
praveen-garg: exit if NOT Red Hat|CentOS with `echo "This script is compatible with Red Hat Linux distributions only!"`

PR #437, opened 2019-05-07
FOGL-2790 (patch) readme fixes; yum localinstall should auto install the required dependencies

praveen-garg:


rpm-readme.md
```diff
@@ -56,19 +56,39 @@ Done.
...
+```
+
+> You can also use `--replacefiles` instead of `--force`
+
+OR
 
 ```
-$sudo cp packages/build/foglamp-gui-x.y.z/RPMS/x86_64/foglamp-gui-x.y.z-1.x86_64.rpm /var/cache/yum/x86_64/.
-$sudo rpm  -i --force /var/cache/yum/x86_64/foglamp-gui-x.y.z.x86_64.rpm
+$ sudo yum localinstall /var/cache/yum/x86_64/foglamp-gui-x.y.z-1.x86_64.rpm
```
monika-Dianomic: `--force` is not working with `localinstall`, then what should be used instead of `--force` in case of conflicting?

rpm-readme.md
```diff
@@ -56,19 +56,39 @@ Done.
 
 ```
 
-#### Installing rpm package
+> You may want to check RPM package content with `sudo rpm -ql foglamp-gui-x.y.z-1.x86_64.rpm`
```
mshariq-nerd: - Put `-p` option as well to check the content  of uninstalled RPM package. 
e.g. `$ sudo rpm -qplv foglamp-gui-x.y.z-1.x86_64.rpm`

- To check installed package content then use `-q` option only with no package extension (.rpm)
e.g. `sudo rpm -ql foglamp-gui-x.y.z-1.x86_64`

options
`-q` - to specify it as a query command
`-l` - to list the files in the package
`-p` -  to query the uninstalled package file.
`-v` - (verbose) to provides additional information (permissions, owner, etc.)

praveen-garg: Fixed!

rpm-readme.md
praveen-garg: Need to figure it out. should I add notes about it as TODO/ FIXME:? And request to use `sudo rpm  -i --force ...` until we fix?

monika-Dianomic: Yes, we should add it as TODO.

PR #440, opened 2019-05-10
FOGL-2797: Fixed grab git version info on running e2e tests

mshariq-nerd:


git-version
```diff
@@ -0,0 +1,15 @@
...
+
+v=$(git --version)
+echo $v
+if [[ $v == "git version 2"* ]]; then
+    echo "Found git version > 2.x"
+    node grab-git-info
+else
+    echo "Found git version < 2.x"
+    if [ ! -f "src/git-version.json" ]; then
+    echo > ${src}/git-version.json
```
praveen-garg: indent please

praveen-garg: what is the `${src}` here?

git-version
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+v=$(git --version)
+echo $v
+if [[ $v == "git version 2"* ]]; then
+    echo "Found git version > 2.x"
+    node grab-git-info
+else
+    echo "Found git version < 2.x"
```
praveen-garg: May be git is not available at all; remove this echo.  We already have `echo $v` on top

PR #468, opened 2019-06-25
FOGL-2927: Respect the existence of "readonly" key with value true/false on GUI

mshariq-nerd:


src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
```diff
@@ -63,7 +63,7 @@ export class ViewConfigItemComponent implements OnInit, OnChanges {
         });
         // check if editable config item found, based on readonly property
         for (const el of this.categoryConfiguration.value) {
-          if (!has(el, 'readonly')) {
+          if (!has(el, 'readonly') || el.readonly === 'false') {
```
praveen-garg: I assume it will be always lowercase or we can compare el.readonly.lowercase to allow False, FALSE as well

PR #472, opened 2019-07-03
FOGL-2911: Asset graph space usage

mshariq-nerd:
1. Allowed toggle between summary and graph.
2. Removed dots in graph lines

Screen Shot 2019-07-03 at 5 31 24 PM

Screen Shot 2019-07-03 at 5 32 41 PM

### DIFF
--- a/src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
+++ b/src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
@@ -36,7 +36,9 @@ export class ReadingsGraphComponent implements OnDestroy {
   public isSpectrum = false;
   public polyGraphData = {};
   public showSummary = false;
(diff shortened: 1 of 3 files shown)


src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
```diff
@@ -35,6 +35,8 @@ export class ReadingsGraphComponent implements OnDestroy {
   public showSummarySpinner = true;
   public isSpectrum = false;
   public polyGraphData = {};
+  public showSummary = false;
+  public toggleSummaryGraphButtonText = 'Show Summary';
```
praveen-garg: it would be better to have `const SHOW_SUMMARY_TEXT =  'Show Summary'` and `const SHOW_GRAPH_TEXT =  'Show Graph' and use them to change at single place.

PR #477, opened 2019-07-09
[FOGL-2973]: Display help text for Trigger and Enabled fields in Notification

monika-Dianomic:
trigger-tooltip


src/app/components/core/notifications/add-notification-wizard/add-notification-wizard.component.html
```diff
@@ -167,7 +167,7 @@
         <div class="box step-content" id="c-6">
           <div class="field is-horizontal">
             <div class="field-label">
-              <label class="label">Trigger</label>
+              <label class="label has-text-left tooltip is-tooltip-right is-tooltip-multiline" data-tooltip="One Shot: only trigger once while in triggered state. Re-Triggered: trigger at every polling while in triggered state. Toggled: trigger once when triggered, and again when cleared.">Trigger</label>
```
praveen-garg: it would be great if we can have line breaks; so that it appears like:

```
One Shot: ...
Re-Triggered: ...
Toggled: ...
```

monika-Dianomic: I have tried to do the same but could not get success to insert html (to line break) in tooltip text.

billehunt: I agree with Praveen that line breaks would be better, but this is great as is.

PR #481, opened 2019-07-15
FOGL-2984: Changed csv formate for single asset reading download from asset page

mshariq-nerd:


src/app/services/generate-csv.service.ts
```diff
@@ -50,4 +55,27 @@ export class GenerateCsvService {
...
   }
+
+  ConvertAssetReadsToCSV(assetData: any) {
+    let str = '';
+    let row = 'timestamp';
+    const keys = Object.keys(assetData[0].reading);
+    for (const header of keys) {
+      row += ',' + header;
+    }
+    row = row.slice(0, -1);
```
praveen-garg: remove `slice(0, -1);` as row has default value  timestamp and `row += ',' + header;`

praveen-garg:
I think - we can hard code the check for `spectrum`, as we do for 3d graph.

PR #482, opened 2019-07-16
FOGL-2933: Add support for password configuration items

mshariq-nerd:
FOGL-2933 - Fixed  Point:3
> Two input files should be created always, an X and X confirm. On save these should be checked to confirm they are the same. Only one is sent to the API. An error is generated if they differ.

Screen Shot 2019-07-23 at 1 17 41 PM

Screen Shot 2019-07-23 at 1 17 03 PM

Screen Shot 2019-07-23 at 1 17 22 PM

### DIFF
--- a/src/app/components/core/configuration-manager/view-config-item/view-config-item.component.html
+++ b/src/app/components/core/configuration-manager/view-config-item/view-config-item.component.html
@@ -25,10 +25,20 @@
                       [ngModel]="setConfigValue(configItem.value)" [value]='setConfigValue(configItem.value)'
                       [attr.maxLength]="configItem?.value?.length" />
 
-                    <input *ngIf="getConfigAttributeType(configItem.value.type) === 'PASSWORD'"
-                      name="{{configItem.value.key}}" class="input is-fullwidth is-small" type="password"
-                      [ngModel]="setConfigValue(configItem.value)" [value]='setConfigValue(configItem.value)'
-                      [attr.maxLength]="configItem?.value?.length" />
+                    <div class="field has-addons" *ngIf="getConfigAttributeType(configItem.value.type) === 'PASSWORD'">
+                      <div class="control">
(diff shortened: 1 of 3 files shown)


src/app/components/core/configuration-manager/view-config-item/view-config-item.component.html
```diff
@@ -25,10 +25,38 @@
...
                       [attr.maxLength]="configItem?.value?.length" />
 
-                    <input *ngIf="getConfigAttributeType(configItem.value.type) === 'PASSWORD'"
-                      name="{{configItem.value.key}}" class="input is-fullwidth is-small" type="password"
-                      [ngModel]="setConfigValue(configItem.value)" [value]='setConfigValue(configItem.value)'
-                      [attr.maxLength]="configItem?.value?.length" />
+                    <div class="field has-addons" *ngIf="getConfigAttributeType(configItem.value.type) === 'PASSWORD'">
+                      <div class="control">
+                        <input #pwd (input)="showConfirmPassword()" name="{{configItem.value.key}}"
+                          class="input is-fullwidth is-small" type="password" onPaste="return false"
```
praveen-garg: placeholder="password"

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.html
```diff
@@ -25,10 +25,38 @@
...
+                          </span>
+                        </a>
+                      </div>
+                    </div>
+                    <div class="field"
+                      *ngIf="getConfigAttributeType(configItem.value.type) === 'PASSWORD' && passwordOnChangeFired">
+                      <div class="control">
+                        <div id="confirm-password" class="field has-addons">
+                          <div class="control">
+                            <input #confirmPwd name="confirm-password" class="input is-fullwidth is-small"
```
praveen-garg: `padding-right: 2rem;` or `padding-right: 33px;`

PR #483, opened 2019-07-19
FOGL-2982 : Add ability to edit python file

mshariq-nerd:
> Increased the modal size to 80% of the page width.

1) South service modal
south-service-modal

2) Notification wizard
wizard-modal

3) Notification modal
notification-modal

### DIFF
--- a/src/styles.css
+++ b/src/styles.css
@@ -176,12 +176,6 @@ table.scroll tbody {
   }
 }
 
-@media screen and (min-width: 769px) {
-  .modal-card {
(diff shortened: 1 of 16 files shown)


src/styles.css
```diff
@@ -176,6 +176,12 @@ table.scroll tbody {
   }
 }
 
+@media screen and (min-width: 769px) {
```
praveen-garg: May be better to keep it components specific. South/ North/ Notify/ View Config Item components etc.

PR #505, opened 2019-08-30
FOGL-3232: Notification Service not available immediately after enabling (1.7.0 RC)

mshariq-nerd:


src/app/components/core/notifications/notifications.component.ts
```diff
@@ -157,11 +126,10 @@ export class NotificationsComponent implements OnInit {
     this.servicesApiService.addService(payload)
       .subscribe(
         () => {
-          /** request done */
-          this.ngProgress.done();
           this.alertService.success('Notification service added successfully.', true);
           this.isNotificationServiceAvailable = true;
           this.isNotificationServiceEnabled = true;
```
praveen-garg: Please remove this; and let it be set with `checkServiceStatus`

PR #516, opened 2019-09-16
FOGL-3268: Codemirror editor for config items of type script (python) and JSON

MonikaSharma06:
1) Codemirror editor added on wizard and  link modal page of South, North, Notification and Filter Component.
2) Option to choose editor theme (light/dark) on Settings page; **default is light**

Codemirror_dark_theme_script
Codemirror_dark_theme_json_with_lint

Setting_page

Codemirror_light_theme_script

Codemirror_light_theme_json_with_lint


praveen-garg:
Can we do something for the border of editor box in case of light theme?

docs/developers-guide.md
```diff
@@ -97,7 +97,6 @@ protractor-beautiful-reporter |  1.2.8       |  1.2.8                | MIT
 node-git-describe | 4.0.4     | 4.0.4        | MIT
 angular-plotly.js | 1.3.2     | 1.3.2        | MIT
 plotly.js         | 1.48.1    | 1.48.1       | MIT
-jsoneditor        | 6.2.0     | 6.2.0        | Apache-2.0
```
shariq1501, line 104: Add codemirror info

src/app/app.component.ts
```diff
@@ -5,6 +5,17 @@ import { SidebarModule } from 'ng-sidebar';
...
+import '../../node_modules/codemirror/mode/javascript/javascript';
+import '../../node_modules/codemirror/mode/python/python';
+import '../../node_modules/codemirror/addon/fold/foldgutter';
+import '../../node_modules/codemirror/addon/fold/brace-fold';
+import '../../node_modules/codemirror/lib/codemirror';
+import '../../node_modules/codemirror/addon/edit/closebrackets';
+import '../../node_modules/codemirror/addon/edit/matchbrackets';
+import '../../node_modules/codemirror/addon/lint/lint';
+import '../../node_modules/codemirror/addon/lint/javascript-lint';
+import '../../node_modules/codemirror/addon/lint/json-lint';
```
shariq1501: Not sure if this is the correct place to put all these imports here. We always put all `node_module` file in angular.json.

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.html
```diff
@@ -72,20 +72,25 @@
                       appNumberOnly />
 
                     <ng-container *ngIf="getConfigAttributeType(configItem.value.type) === 'LONG_TEXT'">
-                      <json-editor #editor [options]="editorOptions" [data]="configItem?.value?.value != undefined ? configItem?.value?.value : configItem?.value?.default">
-                      </json-editor>
-                      <input class="input" type="hidden" name="{{configItem.value.key}}" [ngModel]="editor.getText()"
-                        [jsonValue]="setConfigValue(configItem.value)" />
+                      <ngx-codemirror #jsoneditor name="{{configItem.value.key}}" [options]="{theme: selectedTheme,mode: 'application/json',
+                        lineNumbers: true,lineWrapping: true,foldGutter: true,gutters: ['CodeMirror-linenumbers', 'CodeMirror-foldgutter', 'CodeMirror-lint-markers'],
+                        autoCloseBrackets: true,matchBrackets: true,lint: true}" [autoFocus]="true" [ngModel]="setConfigValue(configItem.value)" (ngModelChange)="checkValidJson(jsoneditor.value)">
```
shariq1501: can't we put all this configuration 
`{theme: selectedTheme,mode: 'application/json', lineNumbers: true,lineWrapping: true,foldGutter: true,gutters: ['CodeMirror-linenumbers', 'CodeMirror-foldgutter', 'CodeMirror-lint-markers'], autoCloseBrackets: true,matchBrackets: true,lint: true}`
in an object inside component `ts` file? And use that object here in html.

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.html
```diff
@@ -72,20 +72,25 @@
...
+                      </p>
                     </ng-container>
 
                     <div *ngIf="getConfigAttributeType(configItem.value.type) === 'SCRIPT'">
-                      <textarea #textarea rows="10" class="textarea is-fullwidth is-small"
-                        name="{{configItem.value.key}}" [ngModel]="fileContent || setConfigValue(configItem.value)"
-                        (click)="getFileName(configItem?.value?.file)"
-                        [value]='fileContent || setConfigValue(configItem.value)' [disabled]="!isFileUploaded">
+                      <ngx-codemirror name="{{configItem.value.key}}" [options]="{theme: selectedTheme,mode: 'text/x-python',
+                      lineNumbers: true,lineWrapping: true,foldGutter: true,gutters: ['CodeMirror-linenumbers', 'CodeMirror-foldgutter', 'CodeMirror-lint-markers'],
```
shariq1501: Move all this configuration inside `ts` file and use that object here html.

src/app/components/core/settings/settings.component.ts
```diff
@@ -20,8 +20,9 @@ export class SettingsComponent implements OnInit {
   pingInterval: string;
   refreshInterval: string;
   serviceUrl = '';
+  selectedTheme = localStorage.getItem('OPTED_THEME');
```
shariq1501: inside `selectedTheme`  on first load there will be empty value. So it should be conditional here either `default` or `stored` value.

src/app/services/shared.service.ts
```diff
@@ -7,4 +7,5 @@ export class SharedService {
   public isAdmin: BehaviorSubject<boolean> = new BehaviorSubject<boolean>(false);
   public isServiceUp: BehaviorSubject<boolean> = new BehaviorSubject<boolean>(false);
   public showLogs: BehaviorSubject<any> = new BehaviorSubject<any>(false);
+  public theme: BehaviorSubject<any> = new BehaviorSubject<any>(localStorage.getItem('OPTED_THEME'));
```
shariq1501: There is no need to read value from localStorage. Put default value here that otherwise it will be blank on first load until no value saved in localStorage. Also it defined the default behaviour only here.

shariq1501 (changes requested):

src/app/app.component.ts
MonikaSharma06: These are not working after putting in `angular.json`.

src/app/components/core/settings/settings.component.ts
shariq1501: Screen Shot 2019-09-17 at 1 25 59 PM

src/app/app.component.ts
shariq1501: Move imports in `view-config-item.component.ts` file only like given below. 

`
import 'codemirror/mode/javascript/javascript';
import 'codemirror/mode/python/python';
import 'codemirror/addon/fold/foldgutter';
import 'codemirror/addon/fold/brace-fold';
import 'codemirror/lib/codemirror';
import 'codemirror/addon/edit/closebrackets';
import 'codemirror/addon/edit/matchbrackets';
import 'codemirror/addon/lint/lint';
import 'codemirror/addon/lint/javascript-lint';
import 'codemirror/addon/lint/json-lint';
`

praveen-garg (dismissed):
LGTM! This looks better than monaco-editor.

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
```diff
@@ -163,7 +203,11 @@ export class ViewConfigItemComponent implements OnChanges {
         this.fileContent = fileReader.result.toString();
       };
       fileReader.readAsText(file);
-
+      const ext = file.name.substr(file.name.lastIndexOf('.') + 1);
+      if (ext !== 'py') {
+        this.isValidExtension = false;
+        this.isFileUploaded = false;
+      }
       this.filesToUpload.push({ [configItem]: file });
```
praveen-garg: this should be else? i.e. when ext is valid only

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
```diff
@@ -104,6 +111,24 @@ export class ViewConfigItemComponent implements OnChanges {
...
+      lineNumbers: true,
+      lineWrapping: true,
+      foldGutter: true,
+      gutters: ['CodeMirror-linenumbers', 'CodeMirror-foldgutter', 'CodeMirror-lint-markers'],
+      autoCloseBrackets: true,
+      matchBrackets: true,
+      lint: true
+    };
+    if (type === 'JSON') {
+        editorOptions.mode = 'application/json';
```
praveen-garg, line 127: we should set completely new object to editorsOptions config? Instead of overriding the python mode only.

praveen-garg:
```mode: 'application/json',
  gutters: ["CodeMirror-lint-markers"],
  styleActiveLine: true,
  lineNumbers: true,
  line: true,
  lint: true,
```

May be just above with json-lint mod

praveen-garg (changes requested):

billehunt (dismissed):
I like it a lot. @MarkRiddoch do you have any concerns with the libraries added or other?

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
MonikaSharma06: There is no need to set new object to editorsOptions config because all options are same for both cases, only mode is different.

praveen-garg (dismissed):

MarkRiddoch (dismissed):
It look good to me. Slight worry over size, but I think we need to decide what our target environment is for the GUI. If it is not running on the same gateway machine then it is not sure a big issue.

docs/developers-guide.md
```diff
@@ -97,7 +97,8 @@ protractor-beautiful-reporter |  1.2.8       |  1.2.8                | MIT
...
+@ctrl/ngx-codemirror | 2.1.0    | 2.1.0     | MIT
+codemirror    | 5.48.4 | 5.48.4 | MIT
```
shariq1501, line 104: Add jsonlint-mod lib as well here.

shariq1501 (dismissed):

MonikaSharma06:
> It look good to me. Slight worry over size, but I think we need to decide what our target environment is for the GUI. If it is not running on the same gateway machine then it is not sure a big issue.

@MarkRiddoch The size of GUI is 6.1M in this PR and on develop, it is 6.7M. So in the respect of size, we are better than develop.

shariq1501 (approved):

PR #517, opened 2019-09-16
FOGL-3262:  Handle installation and enable state check for Notification service

shariq1501:


praveen-garg (dismissed):
looks okay.  if statusCode == 500  in installNotificationService, instead of InternalServerError  show ` Failed to install` else continue to show whaterver the error.statusText is.

praveen-garg (approved):

MonikaSharma06 (approved):

PR #524, opened 2019-10-01
FOGL-3299: Fixed validation check issue on modals

shariq1501:


praveen-garg (dismissed):

src/app/services/validate-form.service.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  constructor() { }
+
+  checkViewConfigItemFormValidity(viewComponent: ViewConfigItemComponent) {
+    if (viewComponent !== undefined) {
+      const isValidPassword = viewComponent.passwordMatched;
+      const isValidJSON = viewComponent.isValidJson;
+      const isValidFileExtension = viewComponent.isValidExtension;
+      const isValidForm = viewComponent.form.valid;
+      return isValidPassword && isValidForm && isValidJSON && isValidFileExtension;
```
praveen-garg: by default these are true? what if any of such item does not exist on `viewComponent`?

shariq1501: By default all are true. There are member variables of viewComponent and will always be available until removed but yes value can be changes true/false

MonikaSharma06 (dismissed):

MonikaSharma06 (approved):

PR #532, opened 2019-10-15
FOGL-3062: Update GUI to use new bucket API for asset graphs

shariq1501:
~Initial implementation for drag to zoom on chart.~
~check attached video for working.
[data.mov.zip](https://github.com/foglamp/foglamp-gui/files/3729326/Screen.Recording.2019-10-15.at.6.23.02.PM.mov.zip)~

1. **Zooming is done via the mouse wheel or via a pinch gesture** for now because of the library limitation. If we need to use zoom-in/zoom-out icons we need to find some alternative path.

2. Panning can be done via the mouse or with a finger in Pan mode. 

3. Drag to previous/future time will not work in Pan mode.

4.  Drag to previous/future time available in Zoom mode.

5. Reset icon moves the graph into default state.

6. Bucket size calculation is based on duration(length) and max bucket limit is 48 hardcoded right now. 
`public caluclateBucketSize(duration) {
    let bucket = Math.round(duration / 10);
    return bucket = bucket === 0 ? 1 : (bucket > 48 ? 48 : bucket);
  }`

**Note:**  Tabular data and 3D graph feature is not available for now with this change. Hence lab script run data visualization is not completely available from API.


package.json
```diff
@@ -26,12 +26,13 @@
...
     "@angular/platform-browser-dynamic": "~8.2.5",
     "@angular/router": "~8.2.5",
-    "@ng-select/ng-select": "^3.0.7",
     "@ctrl/ngx-codemirror": "^2.1.0",
+    "@ng-select/ng-select": "^3.0.7",
     "@ngx-progressbar/core": "^5.3.2",
     "angular-tree-component": "~8.4.0",
     "animate.css": "^3.7.2",
     "chart.js": "~2.8.0",
+    "chartjs-plugin-zoom": "^0.7.4",
```
praveen-garg, line 35: how much build size has increased with it?

src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
```diff
@@ -39,6 +41,7 @@ export class ReadingsGraphComponent implements OnDestroy {
   public polyGraphData: any;
   public timeDropDownOpened = false;
   public isModalOpened = false;
+  public showResetButton = false;
```
praveen-garg: though we don't have any other reset button in this component or others using this, but better to call it `showResetZoomButton`

billehunt: The zoom looks good, but a couple things we also need to do:
* It will need to be able to zoom out as far as is possible, and I'm not sure how this design will do that. I.e. if the default shows 30 minutes of data, this video shows zooming in to say 1 minute of data and that's great. But it will also need to be able to zoom back out to say 1 hour or 8 hours of data.
* Equally important as zooming is panning. In fact, this is more important, as currently, there's no way to see past data in the product. There needs to be a way to graphically drag left/right (or perhaps other action) to move the graph to the past or the future.

shariq1501:
@billehunt  Please check this latest video of zoom facility. Now we can drag the pan left/right and do scroll zoom but selection of pan portion can not be select now because of the limitation of the chart.js zoom package we can use only one way to zoom. 
[zoom.mov.zip](https://github.com/foglamp/foglamp-gui/files/3734168/zoom.mov.zip)

package.json
shariq1501: Build size is very much the same

shariq1501:
Closing this in favour of https://github.com/foglamp/foglamp-gui/pull/551

PR #533, opened 2019-10-18
FOGL-3369 set required attribute if the mandatory optional attribute is set to …

praveen-garg:
…true for config item

* Added for integer, float, string and URL only

Signed-off-by: Praveen Garg <praveen.garg@nerdapplabs.com>


praveen-garg:
@ashish-jabble  can you please verify this for https://github.com/foglamp/FogLAMP/pull/1747

@monika-Dianomic if looks good we can merge with the fields currently its applied to and check / add for all other possible (including) custom fields too.

shariq1501 (approved):

PR #539, opened 2019-10-25
FOGL-3369 (b): Whitespace handling for required field

MonikaSharma06:
- [x] Whitespace handling for required field
- [x] Fixed JSON error message in the case of two JSON fields


src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
```diff
@@ -452,6 +454,9 @@ export class ViewConfigItemComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   checkValidityOnChange(key: string, configValue: string) {
+    if (configValue.trim().length === 0) {
+      this.form.controls[key].setErrors({'required': true});
```
praveen-garg: This is applicable only of config item optional item mandatory is True. Looks we need to have a custom validator overriding  the required validator to handle whitspaces

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.html
```diff
@@ -93,18 +93,15 @@
 
                     <ng-container *ngSwitchCase="'LONG_TEXT'">
                       <ngx-codemirror #jsoneditor name="{{configItem.value.key}}"
-                        [options]="setEditorConfig(configItem.value.type)" [autoFocus]="true"
-                        [ngModel]="JSON.parse(setConfigValue(configItem.value)) | json " (ngModelChange)="checkValidJson(jsoneditor.value)">
+                        [options]="setEditorConfig(configItem.value.type)" [autoFocus]="true" [jsonValue]="isValidJson"
```
praveen-garg, line 95: what is `[jsonValue]` attirbute, looks we are binding it to True and false only?

praveen-garg (changes requested):

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.html
MonikaSharma06: It is custom attribute here, like `required`. And yes, we are binding it to true and false only. This attribute is set to true and false on the basis of valid and invalid input JSON value.
`this.form.controls[key].setErrors({'jsonValue': false});` 
and showing error msg:
```
<span *ngIf="f?.controls[configItem.value.key]?.hasError('jsonValue')" class="help is-danger">
    Invalid JSON
</span>
```

praveen-garg: then better to rename it as `validJSON`

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
```diff
@@ -459,6 +461,9 @@ export class ViewConfigItemComponent implements OnInit, OnChanges, OnDestroy {
             .replace(new RegExp(key, 'g'), `'${configValue}'`);
         }
       }
+      if (configItem.hasOwnProperty('mandatory') && configValue.trim().length === 0) {
```
praveen-garg: we should also check if mandatory is true, otherwise this will trigger even if it is set to false.

praveen-garg (changes requested):

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
praveen-garg: and this will also apply to Advanced Config as well?

shariq1501 (approved):

praveen-garg (approved):

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.html
MonikaSharma06: This is the name of directive, and changing its name effects on many components. So it is better to keep as it is.

src/app/components/core/configuration-manager/view-config-item/view-config-item.component.ts
MonikaSharma06: done

PR #542, opened 2019-11-13
UI fixes for mobile viewport

MonikaSharma06:


src/app/components/core/north/north.component.html
```diff
@@ -8,7 +8,8 @@
           </p>
         </div>
       </div>
-      <a id="add_task" class="fix-pad button is-light" (click)="addNorthInstance()">Create North Instance &nbsp;
```
praveen-garg, line 11: leave it as `Create North Instance`

praveen-garg (commented):
on south, show + only when viewPort is 'mobile' otherwise as we are having today

praveen-garg (dismissed):

praveen-garg (approved):

shariq1501 (approved):

PR #550, opened 2019-12-03
FOGL-3483: Allow menu items to open in new tab

MonikaSharma06:


src/app/app.component.ts
```diff
@@ -30,6 +30,18 @@ export class AppComponent implements OnInit {
   }
 
   ngOnInit() {
+    if (localStorage.getItem('token')) {
```
praveen-garg: can we take these in a list of string, and iterate over it.    
```
const storageItems = ['token', ...]
for si in storageItems {
 ....
}
```

praveen-garg (changes requested):

PR #551, opened 2019-12-05
FOGL-3062: Update GUI to use new bucket API for asset graphs

shariq1501:
**Current Data Mode**
- [x] Zoom In  (based on defined window sized calculation)
- [x] Zoom Out  (based on defined window sized calculation) 

**Historic Data Mode (Pan Mode)**
- [x] Pan to the Past (Drag Right) 
- [x] ZoomIn
- [x] ZoomOut 

Note: This graph functionality is done using the Plotly.js library that we are already using for 3D graph. Plotly.js implicitly provide this Zoom facility.


src/app/services/assets.service.ts
```diff
@@ -42,6 +42,20 @@ export class AssetsService {
...
+  public getAssetReadingsBucket(payload) {
+    let params = new HttpParams();
+    if (payload.start !== 0) {
+      params = params.append('start', payload.start);
+    }
+    if (payload.len) {
+      params = params.append('length', payload.len);
+    }
+
+    return this.http.get(`${this.GET_ASSET}/${payload.assetCode}/bucket/${payload.bucketSize}`, { params: params }).pipe(
```
MonikaSharma06: `assetCode` should be encoded here.

src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
```diff
@@ -333,195 +218,123 @@ export class ReadingsGraphComponent implements OnDestroy {
...
+    const panDeltaTime = moment.duration(panReleaseTime.diff(panClickTime)).asSeconds();
+    console.log('panDeltaTime', panDeltaTime);
+
+    const now = moment.utc(new Date()).valueOf() / 1000.0; // in milliseconds
+    console.log('now', now);
 
-  keyDescOrder = (a: KeyValue<number, string>, b: KeyValue<number, string>): number => {
-    return a.key > b.key ? -1 : (b.key > a.key ? 1 : 0);
+    const currentTimeWindow = Utils.getTimeWindow(this.timeWindowIndex); // in seconds
+    console.log('current time', currentTimeWindow);
```
praveen-garg: console.log('current timeWindow', currentTimeWindow);

src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
```diff
@@ -333,195 +218,123 @@ export class ReadingsGraphComponent implements OnDestroy {
...
-  isEmptyObject(obj) {
-    return (obj && (Object.keys(obj).length === 0));
-  }
+    // console.log('panClickTime(unix timestamp)', moment(panClickTime.format()).valueOf() / 1000);
+    // console.log('panReleaseTime(unix timestamp)', moment(panReleaseTime.format()).valueOf() / 1000);
+
+    const panDeltaTime = moment.duration(panReleaseTime.diff(panClickTime)).asSeconds();
+    console.log('panDeltaTime', panDeltaTime);
+
+    const now = moment.utc(new Date()).valueOf() / 1000.0; // in milliseconds
```
praveen-garg: with `/ 1000.0` these are now in seconds. i.e. standard epoch time (with ms level precision)

src/app/utils.ts
```diff
@@ -4,10 +4,10 @@ export const GRAPH_REFRESH_INTERVAL = 5000; // milliseconds
 export const STATS_HISTORY_TIME_FILTER = '10'; // minutes
 export const ASSET_READINGS_TIME_FILTER = 600; // seconds
 export const COLOR_CODES = ['#3498DB', '#85C1E9', '#239B56', '#82E0AA', '#B03A2E', '#F1948A', '#FF8C00', '#FF0000',
```
praveen-garg, line 6: I believe, We can remove this now. However for R/G/B (e-pHat) datapoints/ attributes, let's keep matching colors as we do in current develop for legends / graph lines

src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
```diff
@@ -113,201 +157,42 @@ export class ReadingsGraphComponent implements OnDestroy {
     if (activeDropDowns.length > 0) {
       activeDropDowns[0].classList.remove('is-active');
     }
```
MonikaSharma06: This code should be removed.

MarkRiddoch (commented):
I'm seeing some odd results still with the latest version of develop and this branch

As I zoom in I see white space appear on the left hand side of the graph. If I zoom more then this white space area gets bigger.

If I drag left such that the right hand edge would be a time in the future the graph is dragged back so that the right hand edge is now - which is correct. However I am left with white space to the left of the graph.

shariq1501:
@MarkRiddoch I am not able to reproduce the issue mention by you above. Though I have fixed some issues related to graph update. Please check now.
Also could you please check the timezone of the machine where FogLAMP and GUI is running. Make sure FogLAMP and GUI are running in same timezone.

src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
```diff
@@ -1,61 +1,156 @@
...
+          }
+        }
+      }
+    ]]
+  };
+
+  payload = {
+    assetCode: '',
+    start: 0,
+    len: 600,
```
MonikaSharma06: It would be better to put 600, i.e. Max Datapoints, in a variable and use that variable throughout the file.

src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
```diff
@@ -332,194 +249,245 @@ export class ReadingsGraphComponent implements OnDestroy {
...
-    return (obj && (Object.keys(obj).length === 0));
+  public updateTimeWindowText(timeWindowText) {
+    this.layout.xaxis.title.text = `Time Window - ${timeWindowText}`;
   }
 
-  keyDescOrder = (a: KeyValue<number, string>, b: KeyValue<number, string>): number => {
-    return a.key > b.key ? -1 : (b.key > a.key ? 1 : 0);
+  panModeZoom(zoomOut = false) {
+    // <Start> =(previous Start) + (previous Timespan in seconds) / 2 - (new Timespan in seconds) /2
+    console.log('lateset payload', this.payload);
```
MonikaSharma06: lateset -> last OR previous

MarkRiddoch:
Some observations on the pan and zoom functionality

- Performance is poor, due I suspect to the underlying queries taking a long time. However it would help if the GUI indicated it was busy waiting for results as it is not obvious currently and it is easy to click again and send up having lots of stored up clicks before seeing any data back.

- It might be better to start with a smaller view on the data as it takes a long time to get the initial screen

- It is possible to zoom the graph to the future, the really doesn't have any meaning and it would probably aid usability if this was removed.

- I often get left with a blank, unresponsive screen. I guess because it is waiting for data. Can we keep the previous view until the new data is ready? Combine this with busy feedback and it would probably prevent the user clicking wildly, storing up more refreshing and never seeing anything.

As it stands I don't think this is really usable with any volume of data.

On revisiting this is seems that smaller volumes of data work much better. As soon as you have high bandwidth data things become difficult. I suspect this is down to the volume of data and the time the queries are taking. There may be nothing we can do currently with this, it may also be the case that this is actually still true with the old graph mechanism. It is also probably true that most of our current user base do not have high bandwidth data, so maybe we can go with this on those grounds.

PR #554, opened 2019-12-18
[FOGL-3062.tickrange]: Show different tickformats according to data range on x-axis

MonikaSharma06:


src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
```diff
@@ -310,8 +313,35 @@ export class ReadingsGraphComponent implements OnDestroy {
...
+    }
+    // above 6 months
+    if (15552000 <= length) {
+      this.layout.xaxis.tickformat = '%b %y';
+    }
+  }
+
   dragGraph(event: any) {
-    if (event['xaxis.range[0]'] === undefined) {
+    if (event['xaxis.range[0]'] === undefined  || this.numReadings[0] === undefined) {
```
praveen-garg: can this be `this.numReadings` length check?

src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
```diff
@@ -293,13 +293,16 @@ export class ReadingsGraphComponent implements OnDestroy {
...
+    const timeWindow = Utils.getTimeWindow(this.timeWindowIndex);
+    this.updateXAxisTickFormat(timeWindow.value);
   }
 
   public zoomGraph(seconds: number) {
     this.zoom = true;
     const maxDataPoints = 600;
     const bucket = seconds / maxDataPoints;
     const length = seconds;
+    this.updateXAxisTickFormat(length);
```
praveen-garg: when there are many readings in DB, API response may be delayed. do we want to update here (per timeWindow text) or after `getAssetReadings` call.

praveen-garg (approved):

src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
MonikaSharma06: yes, it can be.

src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
MonikaSharma06: Removed from here, now it is updated after `getAssetReadings` call.

shariq1501 (approved):

PR #555, opened 2019-12-19
[FOGL-3062patch]: refactoring in code

MonikaSharma06:


src/app/components/core/asset-readings/readings-graph/readings-graph.component.ts
```diff
@@ -58,7 +59,7 @@ export class ReadingsGraphComponent implements OnDestroy {
     }
   };
 
-  timeWindowIndex = 23;  // initial value is 23rd index i.e 720s
+  timeWindowIndex = this.DEFAULT_TIME_WINDOW_INDEX;  // initial value is 20th index i.e 720s
```
praveen-garg: 600s

src/app/utils.ts
```diff
@@ -130,7 +127,8 @@ export default class Utils {
     };
     const key = Object.keys(timeWindows)[index];
     const value = Object.values(timeWindows)[index];
-    return { key, value };
+    const size = Object.keys(timeWindows).length;
+    return { key, value, size };
```
praveen-garg, line 131: May be we should take const timeWindows out, and instead of returning size  here,  add a function getTimeWindowsSize to return that.

praveen-garg (changes requested):

src/app/utils.ts
MonikaSharma06: It does not make any difference, we can do it in anyway.

praveen-garg: Okay. if @mshariq-nerd is okay with this approach then I'm fine.

praveen-garg (approved):

shariq1501 (approved):

PR #559, opened 2019-12-30
Changed rangemode of graph on dashboard

MonikaSharma06:


src/app/components/core/dashboard/dashboard.component.ts
```diff
@@ -272,6 +276,44 @@ export class DashboardComponent implements OnInit, OnDestroy {
...
+          font: {
+            size: 14,
+            color: '#7f7f7f'
+          }
+        }
+      },
+      yaxis: {
+        fixedrange: true,
+        rangemode: 'nonnegative',
+        range: [0, 8]
```
praveen-garg, line 269: cannot we make it `range: value !== 0 ? [0, 8] : []`

BTW, what is this value here?

src/app/components/core/dashboard/dashboard.component.ts
```diff
@@ -5,6 +5,7 @@ import { takeWhile, takeUntil } from 'rxjs/operators';
 
 import { AlertService, PingService, StatisticsService } from '../../../services';
 import { GRAPH_REFRESH_INTERVAL, STATS_HISTORY_TIME_FILTER } from '../../../utils';
+import { analyze } from 'tern';
```
praveen-garg: from where we got tern package in?

src/app/components/core/dashboard/dashboard.component.ts
MonikaSharma06: > cannot we make it range: value !== 0 ? [0, 8] : []

No, we can't.

> BTW, what is this value here?

This value is nothing but the range of y-axis when there is no value on y-axis (i.e. - [0, 0, 0, 0]), just to make tick difference in non-floating number.
Otherwise, when data on y-axis is [0, 0, 0, 0], it means the range is [0, 1] and because of this range, tick difference will be in floats number.

src/app/components/core/dashboard/dashboard.component.ts
MonikaSharma06: Mistakenly and useless import, removing it..

praveen-garg (approved):

shariq1501 (approved):

PR #563, opened 2020-01-15
Some changes in footer

MonikaSharma06:
Changes in footer:
1) If content is small, show footer at the bottom of the screen
2) If page content is longer than viewport, show footer after scrolling to the bottom

### DIFF
--- a/src/styles.css
+++ b/src/styles.css
@@ -218,7 +218,19 @@ table.scroll tbody {
   }
 }
 
+.container.is-fluid {
+  min-height: 100%;
+  margin-bottom: -142px; 
+}
 
+.container.is-fluid:after {
+  content: "";
+  display: block;
+}
+
+.footer, .container.is-fluid:after {
+  height: 142px; 
+}
 
 /**
 Custom css for bulma dropdown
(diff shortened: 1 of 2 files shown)


src/styles.css
```diff
@@ -218,7 +218,19 @@ table.scroll tbody {
...
+  margin-bottom: -142px; 
+}
 
+.container.is-fluid:after {
+  content: "";
+  display: block;
+}
+
+.footer, .container.is-fluid:after {
+  height: 142px;
```
praveen-garg, line 232: should be 144px

src/styles.css
```diff
@@ -218,7 +218,19 @@ table.scroll tbody {
   }
 }
 
+.container.is-fluid {
+  min-height: 100%;
+  margin-bottom: -142px;
```
praveen-garg, line 223: should be -144px

praveen-garg (changes requested):
footer padding-top should be 7rem or 6rem; and on settings page it should be overridden to -1 of other pages, as that has version info as well, above the copyright text.

